PR #9, opened 2024-05-17
Sync template changes 20240517

bill-moj:


CHANGELOG.md
```diff
@@ -1,53 +1,65 @@
...
 
+**May 10th 2024** – Derive user details from the `authorization_code` "user" token instead of making an API call to `hmpps-manage-users-api`, thereby removing an unnecessary dependency.
+
+PR: [#352](https://github.com/ministryofjustice/hmpps-template-typescript/pull/352)
+
+---
+
 **February 29th 2024** – Use same node version for outdated check and security scan. This currently defaults to node 16
 
-PR: [#321](https://github.com/ministryofjustice/hmpps-challenge-support-intervention-plan-ui/pull/321)
```
iterative-it, line 5: A frustrating part of this merge is it tries to change all the renamed URLs back to the template default. Could you rename them hmpps-template-typescript -> hmpps-challenge-support-intervention-plan-ui? They're in the readme too

bill-moj: readme and changelog are at the moment really just readme and changelog of the template, so the urls with hmpps-template-typescripts would be more correct imo.

i don't have anything to write into the readme yet, we may later write a proper readme and overwrite the content like we do with the -api repo.

as for changelog. can we just keep this sync to the template changelog, and create an alternative changelog file if we need one later?

iterative-it (approved):
LGTM

PR #27, opened 2024-06-20
feat: journey based structure, use jest-dom + supertest for tests

ReedSoftware:


package.json
```diff
@@ -96,7 +81,7 @@
   "dependencies": {
     "@aws-sdk/client-sqs": "^3.596.0",
     "@ministryofjustice/frontend": "^2.1.3",
-    "@opentelemetry/api": "~1.9.0",
+    "@opentelemetry/api": "~1.8.0",
     "agentkeepalive": "^4.5.0",
```
bill-moj, line 85: too bad there's no easy write to put a comment on this to remind me not to update this package.

server/journeys/referral/on-behalf-of/controller.ts
```diff
@@ -0,0 +1,7 @@
+import { Request, Response } from 'express'
+
+export class ReferralOnBehalfOfController {
```
bill-moj, line 3: since we don't have `Referral` in the route name, maybe this can just be `OnBehalfOfController` as well?

server/journeys/referral/view.njk
```diff
@@ -0,0 +1,6 @@
+{% block content %}
+
+  <h1>Referrals is under construction...</h1>
+  <p>Please check back later when there is content to view.</p>
+
+{% endblock %}
```
bill-moj, line 6: better for this to extends from layout.njk?

anyway this is just a place holder at the moment

bill-moj (approved):

package.json
ReedSoftware: I have often wished for the same for similar reasons!

server/journeys/referral/on-behalf-of/controller.ts
iterative-it: It is part of the overall Referral journey so it makes sense to me to have it in the name

package.json
bill-moj: there are suggested workaround such as:
```
    "@opentelemetry/api": "// keep this at 1.8.0 to maintain compatibility with other dependencies",
    "@opentelemetry/api": "~1.8.0",
```

but i wouldn't want to tinker with how this weirdness will interact with renovate and everything else

server/journeys/referral/on-behalf-of/view.njk
```diff
@@ -0,0 +1,7 @@
+{% block content %}
+
+  <h1>Referrals (On Behalf Of) asdadasdwq is under construction...</h1>
+  <p>Please check back later when there is content to view.</p>
+  <b
```
iterative-it, line 5: Incomplete tag here. It'll get fixed when the proper view markup is in though

ReedSoftware: I'm not sure how that happened, apologies and thanks!

server/journeys/referral/on-behalf-of/controller.ts
bill-moj: yep. i'm fine with either:
```
OnBehalfOfController
OnBehalfOfRoutes
```
or:
```
ReferralOnBehalfOfController
ReferralOnBehalfOfRoutes
```

iterative-it (approved):
LGTM

bill-moj (approved):

PR #33, opened 2024-06-24
MIC-189 initial e2e test for Make A Referral journey

bill-moj:


integration_tests/pages/referral/on-behalf-of.ts
```diff
@@ -0,0 +1,9 @@
+import Page, { PageElement } from '../page'
+
+export default class OnBehalfOfPage extends Page {
+  constructor() {
+    super('Referrals (On Behalf Of) asdadasdwq is under construction...')
+  }
+
+  miniProfileDob = (): PageElement => cy.findByText('Date of birth').next('span')
```
iterative-it, line 8: Might be best to have a miniProfile property in the Page class as it's in the main layout. Can be found with `cy.get('[data-qa=mini-profile]')`. We could then assert that the mini profile is displayed on the pages we expect rather than the DOB itself is present. That's better because the contents of the mini profile may be changed by the designers

integration_tests/pages/referral/on-behalf-of.ts
```diff
@@ -0,0 +1,9 @@
+import Page, { PageElement } from '../page'
+
+export default class OnBehalfOfPage extends Page {
+  constructor() {
+    super('Referrals (On Behalf Of) asdadasdwq is under construction...')
+  }
+
+  miniProfileHyperlink = (name: string): PageElement => cy.findByRole('link', { name: new RegExp(name, 'i') })
```
iterative-it, line 8: I think this should be in the base `Page` as the mini profile is likely to be on the majority of pages

iterative-it (approved):
LGTM. Small comment that can be addressed in a later PR

PR #34, opened 2024-06-24
feat: validation 

ReedSoftware:


server/app.ts
```diff
@@ -42,6 +43,7 @@ export default function createApp(services: Services): express.Application {
   app.use(setUpAuthentication())
   app.use(authorisationMiddleware())
   app.use(setUpCsrf())
+  app.use(populateValidationErrors())
```
bill-moj, line 46: unimportant: i think this can go after the prisoner image route, we don't need this middleware for that route.

server/views/partials/formErrorSummary.njk
```diff
@@ -0,0 +1,14 @@
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% set errorSummaryList = validationErrors | buildErrorSummaryList %}
+
+{% if errorSummaryList.length > 0 %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            {{ govukErrorSummary({
+                titleText: "There is a problem",
+                errorList: errorSummaryList,
```
bill-moj, line 9: same title "There is a problem" even when there are multiple errors?

ReedSoftware: Hm, this is how AM does it - I'm not sure what the standard is, but it looks like a lot of other MOJ projects use the same sentence.

server/app.ts
ReedSoftware: Done :)

bill-moj (approved):

PR #36, opened 2024-06-25
MIC-86 referral area-of-work screen

bill-moj:


server/journeys/referral/area-of-work/controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  GET = async (req: Request, res: Response): Promise<void> => {
+    req.journeyData.referral ??= {}
+
+    const areaOfWorkOptions = [
+      {
+        value: '',
+        text: 'Select area',
+        selected: !req.journeyData.referral.refererArea,
+      },
+      ...(await this.csipApiService.getReferenceData(req, 'area-of-work')).map(refData => ({
```
iterative-it, line 16: Would be nice to move all of this down into the service layer so you end up with only `await this.csipApiService.getReferenceData(req, 'area-of-work', 'Select area'))` call with all the mapping and the insertion of the default option done in the service. That way that can be unit tested in isolation

bill-moj: we still have use for ReferenceData[], while this function is generating govukSelect's option items which is slightly different.
i'd prefer to make it a base controller class function to convert ReferenceData[] to govukSelect options.

iterative-it: As long as this code is shared somewhere, it'd fine. Base controller class function makes sense to me

server/services/csipApi/csipApiClient.ts
```diff
@@ -0,0 +1,17 @@
+import RestClient from '../../data/restClient'
+import config from '../../config'
+import { ReferenceData, ReferenceDataType } from '../../@types/csip/csipApiTypes'
+
+export default class CsipApiClient {
+  private readonly restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('Prison API', config.apis.csipApi, token)
```
iterative-it, line 9: Tiny copy paste error, CSIP API rather than Prison API

bill-moj: fixed

server/journeys/referral/area-of-work/schemas.ts
```diff
@@ -0,0 +1,20 @@
...
+  async (req: Request): Promise<z.ZodTypeAny> => {
+    const areaRefDataMap = new Map(
+      (await csipApiService.getReferenceData(req, 'area-of-work')).map(itm => [itm.code, itm]),
+    )
+
+    return z.object({
+      areaOfWork: z
+        .string({ message: 'Select your area of work' })
+        .min(1, 'Select your area of work')
+        .max(12, 'Select your area of work')
```
iterative-it, line 16: I don't think the min max restrictions are needed any more as the refine bit below should cover it?

server/config.ts
```diff
@@ -89,13 +89,21 @@ export default {
...
+        response: Number(get('CSIP_API_TIMEOUT_RESPONSE', 20000)),
+        deadline: Number(get('CSIP_API_TIMEOUT_DEADLINE', 20000)),
+      },
+      agent: new AgentConfig(Number(get('CSIP_API_TIMEOUT_RESPONSE', 20000))),
+    },
     prisonerSearchApi: {
       url: get('PRISONER_SEARCH_API_URL', 'http://localhost:8082', requiredInProduction),
       timeout: {
         response: Number(get('PRISONER_SEARCH_API_TIMEOUT_SECONDS', 20000)),
-        deadline: Number(get('PRISONER_SEARCH_API_TIMEOUT_SECONDS', 20000)),
```
iterative-it, line 96: (Minor) all the others use `_RESPONSE` not `_SECONDS` so `PRISONER_SEARCH_API_TIMEOUT_RESPONSE` especially as the value isn't seconds

server/journeys/referral/area-of-work/schemas.ts
```diff
@@ -0,0 +1,20 @@
...
+        .refine(val => areaRefDataMap.has(val), 'Select your area of work')
```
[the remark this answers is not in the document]
bill-moj: make sense. likely for the other screens to take a similar approach of using one generic error message for all kind of validation errors?

server/journeys/referral/area-of-work/tests.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    const result = await request(appWithError)
+      .get(`/${uuid}/referral/area-of-work`)
+      .expect(200)
+      .expect('Content-Type', /html/)
+
+    const html = createTestHtmlElement(result.text)
+    expect(getByText(html, 'Which area do you work in?')).toBeVisible()
+    expect(getByText(html, 'There is a problem')).toBeVisible()
+    expect(getByText(html, 'Select your area of work')).toBeVisible()
```
iterative-it, line 77: I liked James' `const error = getByRole(topLevelElement, 'link', { name: onBehalfOfError }) expect(error).toBeVisible()` test for this as it asserts that the error summary links to the correct form element

iterative-it (approved):
LGTM

PR #38, opened 2024-06-27
feat: on-behalf-of and start

ReedSoftware:
https://github.com/ministryofjustice/hmpps-challenge-support-intervention-plan-ui/assets/109068236/9a96da16-502f-4523-bd80-aa47329246ff


server/journeys/referral/on-behalf-of/tests.test.ts
```diff
@@ -1,13 +1,24 @@
 import { Express } from 'express'
 import { v4 as uuidv4 } from 'uuid'
 import request from 'supertest'
-import { getByRole } from '@testing-library/dom'
+import { findByText, getByRole, getByText, queryByAttribute } from '@testing-library/dom'
```
bill-moj, line 4: `import { agent as request } from 'supertest'` can help avoid lint error

ReedSoftware: I'm not having any linting errors here? :)

server/journeys/start/controller.ts
```diff
@@ -9,8 +10,21 @@ export class StartJourneyController {
...
+        .getPrisonerDetails(req, prisonerNumber as string)
+        .catch((error: unknown) => {
+          if (!(error instanceof Error)) {
+            throw error
+          }
+          const sanitisedError = error as SanitisedError
+          if (sanitisedError.status === 404) {
+            res.redirect(`${res.locals.digitalPrisonServicesUrl}/prisoner-search`)
+            return
+          }
```
bill-moj, line 23: do we need to restrict it to 404 error only?
what if it fails a regex test and get a 400 bad request error?

ReedSoftware: It's a good point and something we need to decide for all our calls like this - the ticket spec was just on 404s, but worth thinking about this more broadly.

server/journeys/referral/routes.ts
```diff
@@ -4,26 +4,29 @@ import { ReferralRootController } from './controller'
...
 
-export default function routes(csipApiService: CsipApiService): Router {
+export default function routes(csipApiService: CsipApiService, prisonerSearchService: PrisonerSearchService): Router {
   const router = Router({ mergeParams: true })
 
-  router.use('/', Routes())
-
-  router.use('/on-behalf-of', OnBehalfOfRoutes())
-  router.use('/area-of-work', ReferralAreaOfWorkRoutes(csipApiService))
+  router.use('/', StartJourneyRoutes(prisonerSearchService))
```
bill-moj, line 28: i was thinking `/start/routes.ts` can be a base class to be inherited, so here we can make a call like:
```
router.get('/prisoners/:prisonerNumber/referral/start', this.redirectWithPrisonerData('/referral/on-behalf-of'))
```

then we just have one unit test for the base class instead of repeating it for every journey.

but then we only have one journey at the moment, so it doesn't matter.

server/journeys/start/controller.ts
bill-moj: redirecting back to DPS seems like a catch-all handling to me, hence i feel that it should cover all 4XX 5XX errors.
it's probably very rare for such edge cases to happen tho.

server/journeys/referral/routes.ts
ReedSoftware: I think it would end up being more confusing and less obvious what's going on having hidden routes in parent classes, but perhaps let's see where we end up after some more journeys?

We would need a unit test for each journey anyway to ensure that the start route redirects to the specific journey route, imo

server/journeys/start/controller.ts
ReedSoftware: It would be perhaps jarring for a user to be redirected back to the DPS home screen because of an error the server made (5xx) or that we made in our request/response/setup (400, 401, 403 etc), I think? Maybe it would be worth redirecting the user to a generic error page (we have a very basic one)?

iterative-it: I agree with error page for non 404s. The 400 case, while a good suggestion, would only happen if the user entered a manual URL. For the 5XXs we'd want to show the user a friendly error page

server/middleware/validationMiddleware.ts
```diff
@@ -12,6 +12,13 @@ export const buildErrorSummaryList = (array: fieldErrors) => {
   }))
 }
 
+export const findError = (errors: fieldErrors, fieldName: string) => {
+  if (!errors) {
+    return null
+  }
+  return errors[fieldName]?.[0]
```
iterative-it, line 19: I think this needs to be
```
return {
    text: errors[fieldName]?.[0],
  }
```
as the `errorMessage` nunjucks property expects `{ text: "error message" }`

server/journeys/referral/on-behalf-of/view.njk
```diff
@@ -1,14 +1,48 @@
...
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% set pageTitle = "Make a CSIP referral - Digital Prison Services" %}
 {% extends "../../../views/pages/index.njk" %}
 
 {% block content %}
+<div class="govuk-grid-column-two-thirds">
+  <h1 class="govuk-heading-l">Make a CSIP referral</h1>
+
+  <p>You can use this service to refer {{ prisoner.firstName }} {{ prisoner.lastName }} for a Challenge, Support and Intervention Plan (CSIP)</p>
```
iterative-it, line 11: Should use `{{ (prisoner.firstName + ' ' + prisoner.lastName) | convertToTitleCase }}` to correctly case prisoner name. Note you'll need to add it to nunjucksSetup.ts.

Also has a full stop after (CSIP) in the design

server/journeys/start/controller.ts
ReedSoftware: We actually already have a middleware for this - I've added an extra test to ensure it's going there when we 500 for `getPrisonerDetails`

server/middleware/validationMiddleware.ts
ReedSoftware: Fixed

server/journeys/referral/on-behalf-of/view.njk
ReedSoftware: Fixed

iterative-it (approved):
LGTM

PR #41, opened 2024-07-03
MIC-87 referral/referrer screen

bill-moj:


server/journeys/referral/referrer/schemas.ts
```diff
@@ -0,0 +1,27 @@
...
+    const areaRefDataMap = new Map(
+      (await csipApiService.getReferenceData(req, 'area-of-work')).map(itm => [itm.code, itm]),
+    )
+
+    return z.object({
+      referredBy: z
+        .string({ message: REFERRED_BY_ERROR_MESSAGE })
+        .min(1, REFERRED_BY_ERROR_MESSAGE)
+        .max(240, REFERRED_BY_LENGTH_ERROR_MESSAGE)
+        .refine(val => val.trim().length > 0, REFERRED_BY_ERROR_MESSAGE),
```
iterative-it, line 21: Does this check make the min check above redundant?

server/journeys/referral/referrer/controller.ts
```diff
@@ -0,0 +1,22 @@
...
+      'area-of-work',
+      'Select area',
+      req.journeyData.referral?.refererArea,
+    )
+    const referredBy = req.journeyData.referral?.referredBy
+    res.render('referral/referrer/view', { areaOfWorkOptions, referredBy, backUrl: 'on-behalf-of' })
+  }
+
+  POST = async (req: Request, res: Response): Promise<void> => {
+    req.journeyData.referral ??= {}
```
iterative-it, line 17: Instead of doing the null coalesce, which I'm guessing is due to `strictNullChecks`, we should use `req.journeyData.referral!.refererArea`. We know in the normal flow that it won't be null and if someone has navigated here directly, it's correct that an exception is thrown. This line allows a user to continue the journey from this point with an invalid referral object as it doesn't contain the isOnBehalfOf value.

bill-moj: I do mean to instantiate a referral object if it's null, so that any screen can be reordered to be the first. If there are multi-question child objects, I would do the same null coalesce, the motive is so that we don't have to look through the other screens ' controller to check where the object is to be instantiated.

We will later have the state validation middleware, so it will eventually just behave the same as doing `!` (only perform a null check that will always pass)

iterative-it: `req.journeyData.referral` is instantiated in the `StartJourneyController` which is also where the prisoner data is retrieved. Regardless of page ordering, none of these journey pages are valid unless the journey itself has been started by that controller.

I'm really keen we don't add any code we don't absolutely need. That null coalesce is another logic path that needs testing and it is not necessary in these journeys.

iterative-it (approved):
LGTM

PR #47, opened 2024-07-08
MIC-99 / MIC-150 referral/details screen

bill-moj:


server/journeys/referral/details/tests.test.ts
```diff
@@ -0,0 +1,340 @@
...
+  it('redirect to go back and set validation errors if incident date is in the future', async () => {
+    await request(app)
+      .post(`/${uuid}/referral/details`)
+      .type('form')
+      .send({
+        incidentLocation: 'A',
+        incidentType: 'B',
+        hour: '',
+        minute: '',
+        incidentDate: '25/12/2077',
```
iterative-it, line 278: From a boundary test perspective, it's best to use something like `const tomorrow = addDays(new Date(), 1)` from `date-fns` as that shows the earliest invalid future date

server/utils/datetimeUtils.ts
```diff
@@ -0,0 +1,50 @@
+const DATE_FORMAT_GB = new Intl.DateTimeFormat('en-GB', {
+  year: 'numeric',
+  month: '2-digit',
+  day: '2-digit',
+})
+
+class ParseResult {
```
ReedSoftware, line 7: I think this would be better as a function.

```
const parseResult = (data?: string) => ({
  data,
  success: !!data,
})
```
and then the result type of the functions using it would be clearer and you wouldn't have to go to the class definition to figure out what it is and what it's doing.

iterative-it (approved):
LGTM

PR #55, opened 2024-07-12
MIC-95 / MIC-152 referral/contributory-factors-comments screen

bill-moj:


server/journeys/referral/contributory-factors-comments/controller.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export class ReferralContributoryFactorsCommentsController {
+  GET = async (req: Request, res: Response): Promise<void> => {
+    const contributoryFactorRows = req.journeyData.referral!.contributoryFactors!.map(factor => [
+      {
+        text: factor.factorType.description,
+      },
+      {
+        classes: 'govuk-link',
+        html: `<a href="${escapeHTML(factor.factorType.code.toLowerCase())}-comment">${escapeHTML(factor.comment) ? 'Edit comment' : 'Add comment'}<span class="govuk-visually-hidden"> for ${escapeHTML(factor.factorType.description)} factor</span></a>`,
```
iterative-it, line 12: There's a built in ` | escape` function you can use if you do this in the view rather than from the controller which I would recommend. This is a view concern rather than controller logic. Have a look at the [appointments extra information view](https://github.com/ministryofjustice/hmpps-activities-management/blob/main/server/views/pages/appointments/create-and-edit/appointment-set/extra-information.njk#L38) as that is what this pattern is based on. That also used a table rather than a summary list

server/utils/utils.ts
```diff
@@ -28,3 +28,17 @@ export const initialiseName = (fullName: string | undefined | null): string | nu
   }
   return `${firstName[0]}. ${array.reverse()[0]}`
 }
+
+export const escapeHTML = (str: string | undefined): string | undefined =>
```
iterative-it, line 32: Would definitely prefer we don't write our own escape functions. See comment above. The formatting and escaping should be done in the njk view

iterative-it (approved):
LGTM

PR #56, opened 2024-07-12
MIC-151/100/173 - Involvement

ReedSoftware:


server/journeys/referral/involvement/controller.ts
```diff
@@ -0,0 +1,37 @@
...
+      isProactiveReferral: Boolean(req.journeyData.referral!.isProactiveReferral),
+      staffAssaulted: formResponsesStaffAssaulted || req.journeyData.referral!.staffAssaulted,
+      assaultedStaffName:
+        res.locals.formResponses?.['assaultedStaffName'] || req.journeyData.referral!.assaultedStaffName,
+      backUrl: true,
+    })
+  }
+
+  POST = async (req: Request<unknown, unknown, SchemaType>, res: Response): Promise<void> => {
+    if (req.body.involvementType) {
```
iterative-it, line 29: I don't think they can get here without this being set?

server/journeys/referral/involvement/controller.ts
```diff
@@ -0,0 +1,37 @@
...
+      req.journeyData.referral!.incidentInvolvement = req.body.involvementType
+    }
+    req.journeyData.referral!.staffAssaulted = req.body.staffAssaulted
+    req.journeyData.referral!.assaultedStaffName = req.body.assaultedStaffName
```
iterative-it, line 33: I think `assaultedStaffName` needs setting to null or empty string if `staffAssaulted` is false. That way it sets the correct state if they change their answer

server/journeys/referral/involvement/schemas.ts
```diff
@@ -0,0 +1,50 @@
...
+        }),
+        staffAssaulted: z
+          .string({ message: isStaffAssaultedErrorMsg })
+          .trim()
+          .refine(val => ['false', 'true'].includes(val), { message: isStaffAssaultedErrorMsg }),
+        assaultedStaffName: z.string().max(240, { message: tooLongStaffMemberErrorMsg }),
+      }),
+    )
+    .strict()
+    .refine(val => val.staffAssaulted === 'false' || val.assaultedStaffName?.trim().length > 0, {
```
iterative-it, line 39: I don't think there's a test for this validation message being shown when staff assaulted is true and they have't entered a value

server/journeys/referral/involvement/controller.ts
ReedSoftware: Ah this is a hangover from when I hadn't quite fixed the typing issue I had. Thanks!

server/journeys/referral/involvement/controller.ts
ReedSoftware: It is, in the schema:

```
.transform(val => ({
      ...val,
      staffAssaulted: val.staffAssaulted === 'true',
      assaultedStaffName: val.staffAssaulted === 'true' ? val.assaultedStaffName : null,
    }))
```

server/journeys/referral/involvement/schemas.ts
ReedSoftware: It's part of my suite of `it.each` tests at the end of the test file, it's this entry:
`[{ _csrf: '', assaultedStaffName: '', staffAssaulted: 'true', involvementType: 'VIC' }, ['assaultedStaffName']],`

server/journeys/referral/involvement/schemas.ts
```diff
@@ -0,0 +1,50 @@
...
+            })
+            return z.NEVER
+          }
+          return incidentInvolvementMap.get(val)!
+        }),
+        staffAssaulted: z
+          .string({ message: isStaffAssaultedErrorMsg })
+          .trim()
+          .refine(val => ['false', 'true'].includes(val), { message: isStaffAssaultedErrorMsg }),
+        assaultedStaffName: z.string().max(240, { message: tooLongStaffMemberErrorMsg }),
```
iterative-it, line 35: Do you think it's worth having a boundary test for this max string length? I would normally have tests for all possible validation errors

ReedSoftware: Oops, yes it is, I meant to do that. Adding now!

server/journeys/referral/involvement/controller.ts
ReedSoftware: Fixed.

server/journeys/referral/involvement/schemas.ts
ReedSoftware: Fixed.

PR #62, opened 2024-07-16
feat: e2e test make a referral

ReedSoftware:


integration_tests/e2e/make-a-referral.cy.ts
```diff
@@ -1,20 +1,241 @@
...
+    cy.findByRole('textbox', { name: /add additional information \(optional\)/i }).type('additional info')
+    checkAxeAccessibility()
+    cy.findByRole('button', { name: /continue/i }).click()
+
+    // TODO: We need to check the answers given are correct on CYA page, but it doesnt exist yet
+
+    // We use this for now as CYA page isn't implemented - change this to click back link when it exists
+    cy.url().should('include', '/check-answers')
+    cy.go('back')
+
```
bill-moj, line 130: the /check-answers page doesn't have a Back link.
once you get there, you're locked into CYA page: most of the other pages should just get you back to the CYA page.

/additional-information would be the final page where you can Back all the way to the first screen.

ReedSoftware: That's why I've used cy.go('back') which is the same as pressing the browser back button and what the comment above it is trying to explain :)

integration_tests/mockApis/csipApi.ts
```diff
@@ -0,0 +1,178 @@
...
+        {
+          code: 'B',
+          description: 'AreaB',
+          createdAt: new Date().toISOString(),
+          createdBy: 'foobar',
+        },
+      ],
+    },
+  })
+}
```
bill-moj, line 30: is it necessary to mock every single reference data type?
would it be okay if we just wildcard match /reference-data/<any-ref-data-type>, and return something generic like code: {'CODE1', descrption: 'Description One'}?

ReedSoftware: It will make it much easier to check that the right values are coming through when we go to check your answers and can check that there is the specific description value there

integration_tests/e2e/make-a-referral.cy.ts
bill-moj: what i'm doing in my PR is that, once you've hit the check-answers page, the back url on all the previous pages will be replaced with 'check-answers' (with the exception of the multi-screen sub journey of first 2 screens). so you won't be able to test the back link.

integration_tests/e2e/make-a-referral.cy.ts
```diff
@@ -1,20 +1,241 @@
...
     cy.signIn()
     cy.visit('/prisoners/A1111AA/referral/start')
-    const onBehalfOfPage = Page.verifyOnPage(OnBehalfOfPage)
-    onBehalfOfPage
-      .miniProfileHyperlink('TestName')
-      .should('have.attr', 'href')
-      .should('equal', 'http://localhost:9091/dpshomepage/prisoner/A1111AA')
+
+    cy.findByRole('radio', { name: /no/i }).click()
+    cy.url().should('include', '/on-behalf-of')
```
iterative-it, line 20: Maybe put this line first as I missed the url check for this page initially

integration_tests/e2e/make-a-referral.cy.ts
ReedSoftware: Ah I see ok, I'll change it then

integration_tests/e2e/make-a-referral.cy.ts
```diff
@@ -1,20 +1,243 @@
...
+  cy.findByRole('heading', { name: /what are the contributory factors\?/i }).should('be.visible')
+  cy.findByRole('checkbox', { name: /factor1/i }).click()
+  cy.findByRole('checkbox', { name: /factor3/i }).click()
+  checkAxeAccessibility()
+  cy.findByRole('button', { name: /continue/i }).click()
+
+  cy.url().should('include', '/contributory-factors-comments')
+  cy.findAllByRole('link', { name: /add comment/i }).should('have.length', 2)
+  cy.findAllByRole('link', { name: /add comment/i })
+    .first()
```
bill-moj, line 214: i think you can `findByRole('link', { name: /add comment on factor1 factor/i })` here

ReedSoftware: Ooh thanks I'll try that

bill-moj (approved):
this PR is large enough as it is. i'm happy to see the CYA and confirmation pages covered in a separate PR

PR #70, opened 2024-07-23
MIC-105: csip save and redirect middleware

ReedSoftware:


server/journeys/referral/check-answers/controller.ts
```diff
@@ -20,6 +22,52 @@ export class ReferralCheckAnswersController {
...
+          incidentTypeCode: referral.incidentType!.code,
+          refererAreaCode: referral.refererArea!.code,
+          referredBy: referral.referredBy!,
+          ...(typeof referral.assaultedStaffName === 'string' && { assaultedStaffName: referral.assaultedStaffName }),
+          descriptionOfConcern: referral.descriptionOfConcern!,
+          incidentInvolvementCode: referral.incidentInvolvement!.code,
+          incidentTime: `${referral.incidentTime!}:00`,
+          isProactiveReferral: referral.isProactiveReferral!,
+          // @ts-expect-error asd
+          isSaferCustodyTeamInformed: 'YES',
```
iterative-it, line 47: It's not passing through the right enum value here. I noticed a [TODO about replacing the enum](https://github.com/ministryofjustice/hmpps-challenge-support-intervention-plan-ui/blob/main/server/%40types/csip/csipApiTypes.ts#L7) with the new `OptionalYesNoAnswer`.

In Appointments, we [exposed the enum as a global in nunjucks](https://github.com/ministryofjustice/hmpps-activities-management/blob/main/server/nunjucks/nunjucksSetup.ts#L192) which [allowed its use in the view rather than strings](https://github.com/ministryofjustice/hmpps-activities-management/blob/main/server/views/pages/appointments/create-and-edit/repeat.njk#L33)

ReedSoftware: Ah crap sorry - I meant to do this as a last thing but completely forgot - thank you.

server/journeys/referral/check-answers/controller.ts
```diff
@@ -20,6 +22,51 @@ export class ReferralCheckAnswersController {
...
+  checkSubmitToAPI: RequestHandler = async (req, res, next) => {
+    const prisoner = req.journeyData.prisoner!
+    const referral = req.journeyData.referral!
+    try {
+      await this.createReferral(req, {
+        logCode: prisoner.prisonId,
+        referral: {
+          contributoryFactors: referral.contributoryFactors!.map(factor => ({
+            factorTypeCode: factor.factorType.code,
+            ...(factor.comment !== undefined && { comment: factor.comment }),
```
JackReeveMoJ, line 34: This is kinda complex, but if we're just not setting `comment` on the object if `factor.comment` is undefined, why do we need this check? afaik setting something to undefined is virtually identical to not setting it?

server/@types/csip/csipApiTypes.ts
```diff
@@ -4,4 +4,4 @@ export type ReferenceData = components['schemas']['ReferenceData']
 export type ReferenceDataType = operations['getReferenceData']['parameters']['path']['domain']
 
 // TODO: to be replaced with Enum from CSIP API
```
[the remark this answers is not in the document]
ReedSoftware: Done :)

iterative-it (approved):
LGTM

PR #71, opened 2024-07-23
MIC-298: Started Screen a CSIP referral journey

JackReeveMoJ:


server/journeys/screen/screen/schemas.ts
```diff
@@ -0,0 +1,44 @@
...
+          return z.NEVER
+        }
+        return saferCustodyOutcomeMap.get(val)!
+      }),
+    })
+    .and(
+      csrfSchema.merge(
+        z.object({
+          reasonForDecision: z
+            .string({ message: REASONS_MSG })
```
ReedSoftware, line 36: You can just chain `.trim()` on here, then simply use `.min(1, { message: REASONS_MSG })`

server/journeys/screen/screen/tests.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    })
+  })
+
+  it('should return a 200 on posting valid data and redirect to check answers', async () => {
+    await request(app())
+      .post(`/${uuid}/screen/screen`)
+      .send({ saferCustodyOutcome: 'WFA', reasonForDecision: 'no action needed' })
+      .expect(302)
+      .expect('Location', 'check-answers')
+  })
```
ReedSoftware, line 155: Would be great to get a test of the validation messages appearing on the screen in the different combinations possible - have a look at `server/journeys/referral/involvement/tests.test.ts` (the last test) for an example if you like.

server/@types/express/index.d.ts
```diff
@@ -59,12 +60,30 @@ export type ContributoryFactor = {
...
+type SaferCustodyScreening = Partial<{
+  outcomeType: ReferenceData
+  reasonForDecision: string
+}>
 
 type Investigation = object
 
 type Plan = object
 
+export type CsipRecord = Partial<{
```
iterative-it, line 72: There's a non obvious difference between the hand coded types in this file, which are session data internal to this UI code, and an API response that you're storing here. In this case you can add `export type CsipRecord = components['schemas']['CsipRecord']` to csipApiTypes.ts to use the types autogenerated via the OpenAPI docs then reference `CsipRecord` in `export type JourneyData`, probably after prisoner.

Prisoner is a special case because we only use one endpoint of the Prisoner Search API and only a tiny subset of the Prisoner response. That's why that one is hand coded rather than autogenerated

server/journeys/screen/screen/controller.ts
```diff
@@ -0,0 +1,24 @@
+import { Request, Response } from 'express'
+import { SchemaType } from './schemas'
+import { BaseJourneyController } from '../../base/controller'
+
+export class ScreenController extends BaseJourneyController {
+  GET = async (req: Request, res: Response): Promise<void> => {
+    const saferCustodyOutcomeItems = await this.getReferenceDataOptionsForRadios(
+      req,
+      'outcome-type',
+      req.journeyData.saferCustodyScreening?.outcomeType || '',
```
iterative-it, line 10: I think this needs to be res.locals.formResponses?.['saferCustodyOutcome'] || req.journeyData. saferCustodyScreening?.outcomeType?.code. That way it will use either what the user has entered or what's in session. The former is for when validation fails and the latter is when they either back through the journey or arrive from check your answers

server/journeys/screen/screen/controller.ts
```diff
@@ -0,0 +1,24 @@
...
+    )
+
+    res.render('screen/screen/view', {
+      saferCustodyOutcomeItems,
+      reasonForDecision: req.journeyData.saferCustodyScreening!.reasonForDecision,
+    })
+  }
+
+  POST = async (req: Request<unknown, unknown, SchemaType>, res: Response): Promise<void> => {
+    req.journeyData.saferCustodyScreening!.outcomeType = req.body['saferCustodyOutcome']
```
iterative-it, line 20: You can use the typed `req.body.saferCustodyOutcome` property set via the validation framework here

server/journeys/screen/start/controller.ts
```diff
@@ -0,0 +1,40 @@
...
+    async (req: Request, res: Response): Promise<void> => {
+      const { csipRecordId, journeyId } = req.params
+
+      try {
+        const csip = await this.csipService.getCsipRecord(req, csipRecordId as string)
+        req.journeyData.csipRecord = csip
+        req.journeyData.saferCustodyScreening = {}
+
+        const prisoner = await this.prisonerSearchService.getPrisonerDetails(req, csip.prisonNumber as string)
+        req.journeyData.prisoner = prisoner
```
iterative-it, line 23: Really minor, this can be `req.journeyData.prisoner = await this.prisonerSearchService.getPrisonerDetails(req, csip.prisonNumber as string)` and that avoids intellij complaining about a redundant local variable
image

server/journeys/screen/screen/tests.test.ts
iterative-it: On subtle test that's missing is one that covers one input value being valid and one not. Both values should be retained and replayed back to the user. In this case, I'd selected "No further action" and not entered anything for reasons and it didn't retain my outcome selection. It's due to the `res.locals.formResponses?.['saferCustodyOutcome'] || req.journeyData. saferCustodyScreening?.outcomeType?.code` PR comment above and validation tests would have caught it.

To be fair, I think we have other pages that have similar missing tests and possible bugs

image

server/journeys/screen/start/controller.ts
```diff
@@ -0,0 +1,40 @@
...
+import CsipApiService from '../../../services/csipApi/csipApiService'
+import PrisonerSearchService from '../../../services/prisonerSearch/prisonerSearchService'
+
+export class StartJourneyController {
+  constructor(
+    private readonly csipService: CsipApiService,
+    private readonly prisonerSearchService: PrisonerSearchService,
+  ) {}
+
+  redirectWithCsipData =
```
iterative-it, line 12: This function will be used by pretty much every journey going forward. I'd move it to the shared /journeys/start/controller.ts file and remove this controller

server/journeys/screen/screen/view.njk
```diff
@@ -0,0 +1,40 @@
...
+                classes: "govuk-fieldset__legend--m"
+            }
+        },
+        items: saferCustodyOutcomeItems,
+        errorMessage: validationErrors | findError('saferCustodyOutcome')
+    }) }}
+    
+    {{ govukCharacterCount({
+        label: {
+          html: '<h2 class="govuk-heading-m">Describe the reasons for this decision</h2>'
```
iterative-it, line 26: This should be
```
text: "Describe the reasons for this decision",
classes: "govuk-label--m"
```
Rather than using the html itself. That way the html will update if the component does e.g. extra accessibility

server/journeys/screen/screen/controller.ts
```diff
@@ -0,0 +1,24 @@
...
+  GET = async (req: Request, res: Response): Promise<void> => {
+    const saferCustodyOutcomeItems = await this.getReferenceDataOptionsForRadios(
+      req,
+      'outcome-type',
+      req.journeyData.saferCustodyScreening?.outcomeType || '',
+    )
+
+    res.render('screen/screen/view', {
+      saferCustodyOutcomeItems,
+      reasonForDecision: req.journeyData.saferCustodyScreening!.reasonForDecision,
```
iterative-it, line 15: Similar to above, this needs to be
```
reasonForDecision:
        res.locals.formResponses?.['reasonForDecision'] || req.journeyData.saferCustodyScreening?.reasonForDecision,
```
to retain what the user entered in the event of a validation fail

server/journeys/screen/screen/tests.test.ts
JackReeveMoJ: Good catch. This is now covered by the cypress e2e tests

ReedSoftware (approved):

server/journeys/screen/screen/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+import { BaseJourneyController } from '../../base/controller'
+import { sortAscending } from '../../../utils/utils'
+
+export class ScreenController extends BaseJourneyController {
+  GET = async (req: Request, res: Response): Promise<void> => {
+    const outcomeTypeItems = this.sortRadios(
+      await this.getReferenceDataOptionsForRadios(
+        req,
+        'outcome-type',
+        res.locals.formResponses?.['outcomeType'] || req.journeyData.saferCustodyScreening?.outcomeType || '',
```
iterative-it, line 12: That last `|| ''` isn't needed here. It'll coalesce to undefined which won't match anything in the list anyway

server/journeys/screen/screen/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  POST = async (req: Request<unknown, unknown, SchemaType>, res: Response): Promise<void> => {
+    req.journeyData.saferCustodyScreening!.outcomeType = req.body.outcomeType
+    req.journeyData.saferCustodyScreening!.reasonForDecision = req.body.reasonForDecision
+    res.redirect('check-answers')
+  }
+
+  private sortRadios = (items: { value: string; text: string | undefined }[]) => {
+    // TODO: No need to manually change ACC once ReferenceData DB is updated
```
iterative-it, line 30: Much better not to do this content change here. There's a ticket to update it in the back end and it's the kind of thing that gets forgotten about

iterative-it (approved):
LGTM

PR #73, opened 2024-07-25
MIC-334: fix(details): proactive/reactive content

ReedSoftware:


server/journeys/referral/details/view.njk
```diff
@@ -66,7 +77,7 @@
         }) }}
         {{ govukSelect({
             label: {
-                html: '<h2 class="govuk-heading-m">What was the incident type?</h2>'
+                html: incidentTypeLabel
```
JackReeveMoJ, line 80: Should we be passing this in as text and letting the component do the formatting for us? With `classes: "govuk-label--m"`

ReedSoftware: Good point, done and pushed.

JackReeveMoJ (approved):

PR #90, opened 2024-07-31
MIC-299: CYA screen for Screen journey

JackReeveMoJ:


server/journeys/screen/check-answers/controller.ts
```diff
@@ -0,0 +1,21 @@
+import { Request, RequestHandler, Response } from 'express'
+import { BaseJourneyController } from '../../base/controller'
+
+export class ScreenCheckAnswersController extends BaseJourneyController {
```
bill-moj, line 4: it's not making use of anything in `BaseJourneyController`
i guess the intent is to either add createScreening function to the base class, or to expose its csipApiService to be used by the child class?

server/journeys/screen/check-answers/tests.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    • Bullet 2
+    • Bullet 3
+    
+    Paragraph
+    
+    <script>alert('concerns');</script>
+    
+    <button>this button should be escaped</button>`
+
+const MOCK_INPUT_TEXT_MULTI_EXPECTED = `Text<br>\n      <br>\n          • Bullet 1<br>\n          • Bullet 2<br>\n          • Bullet 3<br>\n          <br>\n          Paragraph<br>\n          <br>\n          &lt;script&gt;alert('concerns');&lt;/script&gt;<br>\n          <br>\n          &lt;button&gt;this button should be escaped&lt;/button&gt;`
```
bill-moj, line 26: if these test values are re-usable for other tests, we might move them to the test constants file?

bill-moj (approved):

server/journeys/screen/check-answers/tests.test.ts
JackReeveMoJ: I was hoping to do this, glad you think it'll useful too!

server/journeys/screen/check-answers/controller.ts
JackReeveMoJ: Good catch, it will be in use in the next ticket / confirmation for actually sending the data off to the API

bill-moj (dismissed):

bill-moj (approved):

PR #94, opened 2024-08-01
MIC-300: Confirmation screen for Screen flow

JackReeveMoJ:


server/@types/express/index.d.ts
```diff
@@ -29,6 +29,7 @@ export type JourneyData = {
   csipRecord?: CsipRecord
   csipRecordCreated?: boolean
   hasValidationErrors?: boolean
+  journeyCompleted?: boolean
```
bill-moj, line 32: can this new generic `journeyCompleted` property deprecate the `csipRecordCreated` above?

server/journeys/base/controller.ts
```diff
@@ -10,6 +11,9 @@ export class BaseJourneyController {
   createReferral = async (req: Request, createCsipRecordRequest: components['schemas']['CreateCsipRecordRequest']) =>
     this.csipApiService.createReferral(req, createCsipRecordRequest)
 
+  createScreeningOutcome = async (req: Request, payload: Parameters<CsipApiService['createScreeningOutcome']>[1]) =>
+    this.csipApiService.createScreeningOutcome(req, payload)
+
```
bill-moj, line 16: not sure whether it's good to put it in a shared base class, when this will only be ever used by `screen/check-answers/controller`. 
i think an alternative is to make csipApiService public, so the child controller can use the apiservice to make calls directly.

server/journeys/base/controller.ts
```diff
@@ -52,4 +56,12 @@ export class BaseJourneyController {
...
       }))
       .sort((a, b) => sortAscending(a.text, b.text))
+
+  getRecordedByFieldsFromJwt = (token: string) => {
+    const jwt = jwtDecode(token) as { user_name: string; name: string }
+    return {
+      recordedBy: jwt.user_name,
+      recordedByDisplayName: jwt.name,
+    }
+  }
```
bill-moj, line 66: these are decoded and available from `res.locals.user.username` and `res.locals.user.displayName`

server/journeys/screen/check-answers/controller.ts
```diff
@@ -10,8 +11,29 @@ export class ScreenCheckAnswersController extends BaseJourneyController {
...
   }
 
-  checkSubmitToAPI: RequestHandler = async (_req, _res, next) => {
-    // Not implemented yet
+  checkSubmitToAPI: RequestHandler = async (req, res, next) => {
+    const screening = req.journeyData.saferCustodyScreening!
+    try {
+      await this.createScreeningOutcome(req, {
+        outcomeTypeCode: screening.outcomeType!.code,
+        date: new Date().toISOString().substring(0, 10),
```
bill-moj, line 19: i expect this to come up often so i have a `todayString()` util function in datetimeUtils for this

server/journeys/screen/confirmation/view.njk
```diff
@@ -0,0 +1,19 @@
+{% from "govuk/components/panel/macro.njk" import govukPanel %}
+
+{% extends "partials/layout.njk" %}
+
+{% set pageTitle = "Confirmation - Screen CSIP referral" %}
+
+{% set theHtml = ("<div><strong>Decision: </strong>" + outcomeTypeDescription + "</div>") %}
```
bill-moj, line 7: `outcomeTypeDescription | escape`

server/journeys/screen/confirmation/view.njk
```diff
@@ -0,0 +1,19 @@
...
+
+{% block innerContent %}
+    {{ govukPanel({
+        titleText: "CSIP referral screening outcome recorded",
+        html: theHtml
+    }) }}
+    <p class="govuk-body">We've updated the status of the referral to "{{outcomeStatus}}".</p>
+    {% if outcomeTypeDescription == "Progress to investigation" %}
+      <h1 class="govuk-heading-m">What needs to happen next</h1>
+      <p class="govuk-body">An investigation must now be carried out. This should include interviewing {{(prisoner.firstName + " " + prisoner.lastName) | convertToTitleCase | escape}} about the behaviour that led to the referral</p>
```
bill-moj, line 17: content in `{{ }}` are autoescaped, so we don't need to escape here.

if we really want to render html in `{{ }}` we'd need to use `{{ variableWithHtml | safe }}`

server/journeys/base/controller.ts
JackReeveMoJ: Good point, I agree. Worth making this change for createReferral as well?

server/journeys/base/controller.ts
JackReeveMoJ: Excellent, I wasn't aware. This simplifies a lot, thank you

server/journeys/screen/check-answers/controller.ts
JackReeveMoJ: glorious

server/journeys/screen/confirmation/view.njk
JackReeveMoJ: Shouldn't need to escape here because outcomeTypeDescription is only populated by reference data - but it doesn't hurt to add that

server/journeys/base/controller.ts
bill-moj: yea i think so. it's just passing on the function call without performing any logic anyway.

integration_tests/mockApis/componentsApi.ts
```diff
@@ -0,0 +1,17 @@
...
+    request: {
+      method: 'GET',
+      url: '/components/components?component=header&component=footer',
+    },
+    response: {
+      status: 500,
+    },
+  })
+}
+
```
bill-moj, line 14: do we really need this stub?

server/journeys/screen/confirmation/view.njk
```diff
@@ -0,0 +1,19 @@
...
+{% set pageTitle = "Confirmation - Screen CSIP referral" %}
+
+{% set theHtml = ("<div><strong>Decision: </strong>" + outcomeTypeDescription | escape + "</div>") %}
+
+{% block innerContent %}
+    {{ govukPanel({
+        titleText: "CSIP referral screening outcome recorded",
+        html: theHtml
+    }) }}
+    <p class="govuk-body">We've updated the status of the referral to "{{outcomeStatus}}".</p>
```
bill-moj, line 14: it's
```
We’ve updated the status of the referral to “{{outcomeStatus}}”. 
```

not
```
We've updated the status of the referral to "{{outcomeStatus}}".
```

bill-moj: `{{outcomeStatus | sentenceCase(false)}}`

integration_tests/mockApis/componentsApi.ts
JackReeveMoJ: Functionally no, but it clears up the console output during e2e tests and makes debugging easier imo

server/journeys/screen/confirmation/view.njk
JackReeveMoJ: I am very blind to the differences with those unicode characters. Updating the tests took far longer than they should have lol

bill-moj: i just avoid typing and copy and paste lol

bill-moj (approved):

PR #106, opened 2024-08-06
MIC-306: IaR - Investigation: Usual behaviour presentation

JackReeveMoJ:


server/journeys/record-investigation/usual-behaviour-presentation/schemas.ts
```diff
@@ -0,0 +1,14 @@
...
+const REASONS_MSG = "Enter a description of the prisoner's usual behaviour presentation"
+const TOO_LONG_ERROR_MSG = "Description of the prisoner's usual behaviour presentation must be 4,000 characters or less"
+
+export const schema = z.object({
+  personsUsualBehaviour: z
+    .string({ message: REASONS_MSG })
+    .max(4000, TOO_LONG_ERROR_MSG)
+    .trim()
+    .min(1, { message: REASONS_MSG }),
+})
```
bill-moj, line 12: you may want to check with the analyst/designer whether this is okay to `trim` the input.
i think the last answer i got was that we should preserve the user's input as it is, but i'm not very certain about that.

bill-moj (approved):

PR #108, opened 2024-08-06
Task/mic 252 improve code consistency

bill-moj:


server/journeys/referral/safer-custody/schemas.ts
```diff
@@ -1,11 +1,14 @@
 import z from 'zod'
+import { createSchema } from '../../../middleware/validationMiddleware'
 
 const ERROR_MSG = "Select if the Safer Custody team is already aware of this referral or not, or select 'I don't know"
 
-export const schema = z.object({
-  isSaferCustodyTeamInformed: z.enum(['YES', 'NO', 'DO_NOT_KNOW'] as const, {
-    message: ERROR_MSG,
-  }),
+export const YES_NO_ANSWER = z.enum(['YES', 'NO', 'DO_NOT_KNOW'] as const, {
```
JackReeveMoJ, line 6: This enum could potentially be moved into validation utils? YES, NO, DO NOT KNOW must be quite common

server/journeys/screen/confirmation/controller.ts
```diff
@@ -11,6 +11,7 @@ export class ConfirmationController {
   }
 
   private mapOutcomeTypeToStatus(outcomeType: ReferenceData) {
+    // TODO: read computed CSIP status from backend instead
```
JackReeveMoJ, line 14: Might need clarification on this but I thought the decision was made to keep these separate (and the backend had more status types than we show here)

bill-moj: i think you're right that the backend will have more status types than covered by this controller. but the logic should be the same. it's just that backend will cover status types applicable to further down the workflow. up to Screen a Referral, they should have the same logic.

server/journeys/record-investigation/routes.ts
```diff
@@ -23,11 +20,11 @@ function Routes(_csipApiService: CsipApiService): Router {
   return router
 }
 
-export default function routes(csipApiService: CsipApiService, prisonerSearchService: PrisonerSearchService): Router {
-  const router = Router({ mergeParams: true })
+export default function routes({ services, path }: { services: Services; path: string }) {
```
ReedSoftware, line 23: I'm not sure I like this change - it goes from a function signature that specifies two services, to one that could specify any number in the types allowed - how is one supposed to know which to pass? Unless I've missed something?

server/journeys/referral/contributory-factors/schemas.ts
```diff
@@ -1,30 +1,43 @@
...
+export const schemaFactory = (csipApiService: CsipApiService) => async (req: Request) => {
+  const refDataMap = new Map(
+    (await csipApiService.getReferenceData(req, 'contributory-factor-type')).map(itm => [itm.code, itm]),
+  )
+  return createSchema({
+    contributoryFactors: z.union(
+      [
+        z.string().transform<ContributoryFactor[]>((val, ctx) => {
+          if (!refDataMap.has(val)) {
+            ctx.addIssue({
```
ReedSoftware, line 19: Does `transform` do the same as superRefine, whereby if you call `addIssue`, it detects it and throws an issue out?

ReedSoftware: Also, why the change from two separated refine then transform lines, to one transform block?

server/middleware/validationMiddleware.ts
```diff
@@ -21,6 +21,30 @@ export const findError = (errors: fieldErrors, fieldName: string) => {
   }
 }
 
+export const createSchema = <T = object>(shape: T) => zodAlwaysRefine(zObjectStrict(shape))
+
+const zObjectStrict = <T = object>(shape: T) => z.object({ _csrf: z.string().optional(), ...shape }).strict()
+
+const zodAlwaysRefine = <T extends z.ZodTypeAny>(zodType: T) =>
```
ReedSoftware, line 28: Can you explain these changes? I'm not sure I understand them fully, apologies :)

server/journeys/referral/safer-custody/schemas.ts
bill-moj: this zod shape contains an error message specific to this screen.. maybe refactor this later when we have a 2nd YES NO DO NOT KNOW answer later.

server/journeys/record-investigation/evidence-secured/schemas.ts
```diff
@@ -1,14 +1,14 @@
...
 const TOO_LONG_ERROR_MSG = 'Description of the evidence secured must be 4,000 characters or less'
 
-export const schema = z.object({
+export const schema = createSchema({
   evidenceSecured: z
     .string({ message: REASONS_MSG })
     .max(4000, TOO_LONG_ERROR_MSG)
-    .trim()
-    .min(1, { message: REASONS_MSG }),
+    .refine(val => val?.trim().length > 0, { message: REASONS_MSG }),
```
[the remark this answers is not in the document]
bill-moj: we're not transforming the user input.
if they want leading and following whitespaces in their input, we keep them.
trim() would transform the output value rather than just trimming it for validation.

server/journeys/record-investigation/routes.ts
bill-moj: these routes are just passing along the services without making use of them itself, i don't think it's very useful to spell out each service when we can pass them in bulk.
the child routes that actually make use of the services can expand it to:
```
function Routes({ csipApiService }: Services)
```

that wasn't done in this particular file because there is no usage of csipApiService yet.

server/journeys/record-investigation/evidence-secured/schemas.ts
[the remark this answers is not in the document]
ReedSoftware: Oh, I thought we specifically didn't want those whitespaces.

server/journeys/referral/contributory-factors/schemas.ts
bill-moj: failing refine does _not_ prevent the transform getting called.
which means if i separate them, i can't just put a `!` on transform result.
in order to get the type right, i have to do the validation in the transform block and return a z.NEVER.

and yes, transform can do what superRefine does. the main difference is that transform's return value will be used for the result value.

server/middleware/validationMiddleware.ts
bill-moj: zObjectStrict is just a helper function to add _csrf and strict restriction because these two are applicable to all schema.

zodAlwaysRefine. that is to resolve the problem about multi-field validation. we do multi-field validation with a refine call on the root object, and the problem we encountered was that when any field in the object fails the validation, then the refine call on the root object won't be trigger, meaning we will miss that validation.
this is a helper function to work around that by doing a safeParse and adding any error to the ctx. if there is no error, the parse result will be used as transform result. if there is error, parse result will be null, and the original input (req.body) will be returned and used for subsequent refine on the root object.

bill-moj: this avoid the need for working out complicated merge/union/intersection/and to build the schema.

server/journeys/record-investigation/evidence-secured/schemas.ts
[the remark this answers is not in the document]
bill-moj: we asked https://mojdt.slack.com/archives/C06H6BHG2J3/p1717769273472779
and there wasn't a decision on trimming whitespaces

server/journeys/record-investigation/routes.ts
ReedSoftware: Ah ok, makes sense now

server/middleware/validationMiddleware.ts
ReedSoftware: Can you show me a place in the code/journey at the moment where this is a problem so I can understand better? As far as I've seen we've been outputting the correct error messages where more than one field fails validation.

server/journeys/referral/contributory-factors/schemas.ts
ReedSoftware: How come this wasn't an issue before - what changed?

server/middleware/validationMiddleware.ts
bill-moj: `server/journeys/referral/details/schemas.ts` is where it was an issue and you have to change the schema structure in #109 

`server/journeys/referral/involvement/schemas.ts` also had challenges with it, hence you had to put them in separate object and then merge them into one schema. 

this helper function makes sure we always only need one object and don't need to merge multi objects to work around the multi field validation issue.

ReedSoftware: Those two were to ensure the validation messages were output in the correct order - in the end it's just a simple `and` to get them in the right order, no?

server/journeys/referral/contributory-factors/schemas.ts
bill-moj: it wasn't an issue because SchemaType inference did not work as expected. the inferred type turned out to be just `any`.
the hidden issue only surfaced when the inferred schematype is fixed.

server/middleware/validationMiddleware.ts
bill-moj: after an `and` you can't `strict()` it. also stuff like merge/and seems to mess up schema type inference.

server/journeys/referral/contributory-factors/schemas.ts
ReedSoftware: image
it seems to be fine for me?

server/middleware/validationMiddleware.ts
bill-moj: it wasn't just ordering.
before your fix on `server/journeys/referral/details/schemas.ts`, if involvementType input fails the validation (no valid option selected from dropdown), then the validation on hour+minute fields will not be executed.

server/journeys/record-investigation/evidence-secured/schemas.ts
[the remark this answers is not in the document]
bill-moj: got confirmation: https://mojdt.slack.com/archives/C06JRCZ5831/p1723032041968469 that we don't trim

server/journeys/referral/contributory-factors/schemas.ts
ReedSoftware: Typechecking definitely turns it up - I went to `involvement/controller.ts` and changed 
`req.journeyData.referral!.incidentInvolvement = req.body.involvementType`
to
`req.journeyData.referral!.incidentInvolvement = req.body.foo`
and then `npm run typecheck` spits out an error

bill-moj: ahhh it wasn't `Awaited` failing the infer, it was `Promise<z.ZodTypeAny>` lol
but `Awaited` doesn't nicely with intellij, so it'd be difficult for me to figure out whether the type works correctly:
image

bill-moj: with Promise<z.ZodTypeAny>:
image

ReedSoftware: So given that you removed that type off the schema functions anyway, shouldn't this be fine apart from IntelliJ being silly? Maybe worth checking for an update to intellij? A quick check of their website says TS language server support is only supposed to work properly on Ultimate or better editions. Then we wouldn't need this extra library.

bill-moj: testing result shows that AsyncReturnType's only advantage over Awaited<ReturnType> is that it helps intellij show the type of SchemaType correctly. it doesn't affect type check.

and typing req with Request<unknown, unknown, SchemaType> can provide a correctly typed req.body even on intellij IDE.

so it's not worth have this extra dependency of type-fest, especially when only an older version work and we can't update it.

bill-moj: intellij version 2024.2 (released this week) just fixed the issue with Awaited. so no issue at all reverting back to Awaited lol

server/middleware/validationMiddleware.ts
ReedSoftware, line 28:
```suggestion
/*
 * Ensure that all parts of the schema get tried and can fail before exiting schema checks - this ensures we don't have to 
 * have complicated schemas if we want to both ensure the order of fields and have all the schema validation run 
*/
const zodAlwaysRefine = <T extends z.ZodTypeAny>(zodType: T) =>
```

bill-moj: added together with link to zod github issue about this for background info

ReedSoftware (approved):

PR #105, opened 2024-08-06
chore: replace hmpps date picker with moj one

ReedSoftware:

PR #109, opened 2024-08-06
feat: make e2e tests for validation of details page

ReedSoftware:


ReedSoftware:
The plan here would be to remove the entirety of the tests.test.ts file as I don't think it's needed anymore at all - all scenarios are covered by the e2e test now

server/journeys/referral/details/schemas.ts
```diff
@@ -61,4 +53,16 @@ export const schemaFactory =
...
+          incidentLocation: z
+            .string({ message: INCIDENT_LOCATION_MSG })
+            .refine(val => incidentLocationMap.has(val), INCIDENT_LOCATION_MSG)
+            .transform(val => incidentLocationMap.get(val)),
+          incidentType: z
+            .string({ message: INCIDENT_TYPE_MSG })
+            .refine(val => incidentTypeMap.has(val), INCIDENT_TYPE_MSG)
+            .transform(val => incidentTypeMap.get(val)),
+        }),
+      )
```
bill-moj, line 67: .and() cannot be followed by `.strict()`,
and `merge` `union` `intersection` doesn't play nicely with inferring schemaType.

hence in PR #108  i'm adding a util class so that we can have all req.body properties defined in a single object, and yet the object.refine().transform() functions will always be executed.

ReedSoftware: Which class or code is that?

ReedSoftware:
This PR is dependent on the fixes in https://github.com/ministryofjustice/hmpps-challenge-support-intervention-plan-ui/pull/105

server/journeys/referral/details/schemas.ts
ReedSoftware: I understand this all better now - I'll wait for #108

server/journeys/referral/details/details.cy.ts
```diff
@@ -151,6 +154,8 @@ context('Make a Referral Journey', () => {
     cy.findByRole('textbox', { name: /minute/i }).type('23')
```
iterative-it, line 154: Might be worth using a different value for hour and minute just to be sure the right one has been retained in each case

ReedSoftware: Done :)

iterative-it (approved):
LGTM

PR #112, opened 2024-08-07
MIC-252 Standardise views.njk files

bill-moj:
MIC-252


JackReeveMoJ (approved):

server/journeys/referral/details/schemas.ts
```diff
@@ -57,23 +57,33 @@ export const schemaFactory = (csipApiService: CsipApiService) => async (req: Req
...
+      ) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: 'Enter a time using the 24-hour clock',
+          path: ['incidentTime'],
+        })
+      }
+    })
+    .transform(val => {
+      const hour = val.hour?.length ? parse24Hour(val.hour) : null
```
ReedSoftware, line 82: Can we do the checks and transformation in `transform`, calling `addIssue` there if needed, instead of having superRefine then transform and doing very similar operations in both?

bill-moj: the helper function does what it says on the tin. it `always refine`. but unfortunately does not `always transform`.

ReedSoftware (approved):

PR #115, opened 2024-08-09
MIC-311 record-investigation/delete-interview/:index screen

bill-moj:
MIC-311

image


server/journeys/record-investigation/delete-interview/tests.cy.ts
```diff
@@ -0,0 +1,64 @@
...
+    cy.contains('dt', 'Comments').next().should('include.text', 'some text')
+
+    cy.findByRole('link', { name: /^back/i })
+      .should('have.attr', 'href')
+      .and('to.match', /\/interviews-summary$/)
+    cy.findByRole('button', { name: /^No, do not delete it/i })
+      .should('have.attr', 'href')
+      .and('to.match', /\/interviews-summary$/)
+
+    proceedToDelete()
```
ReedSoftware, line 44: We should have a test case here to click do not delete

server/journeys/record-investigation/delete-interview/controller.ts
```diff
@@ -0,0 +1,32 @@
...
+        interview: req.journeyData.investigation!.interviews![Number(index) - 1],
+        backUrl: '../interviews-summary',
+      })
+    }
+  }
+
+  POST = async (req: Request, res: Response) => {
+    const { index } = req.params
+    const interviews = req.journeyData.investigation!.interviews!.slice()
+    interviews.splice(Number(index) - 1, 1)
```
JackReeveMoJ, line 23: Do we want any 404 here if the index is not valid or are we happy to just fail. What happens if the user tries `/delete-interview/0` ?

JackReeveMoJ (approved):

PR #116, opened 2024-08-09
MIC-310: IaR - Investigation: Interview details

JackReeveMoJ:


server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+      res.locals.formResponses?.['intervieweeRole'] ||
+        req.journeyData.investigation!.interviews![index]!.intervieweeRole,
+    )
+    return res.render('record-investigation/interview-details/view', {
+      interviewText:
+        res.locals.formResponses?.['interviewText'] || req.journeyData.investigation?.interviews![index]?.interviewText,
+      intervieweeRoleOptions,
+      intervieweeRole:
+        res.locals.formResponses?.['intervieweeRole'] ||
+        req.journeyData.investigation?.interviews![index]?.intervieweeRole,
```
bill-moj, line 29: the pre-selected answer is included in `intervieweeRoleOptions`.
i don't think you need this `intervieweeRole` property?

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  POST = async (req: Request<Record<string, string>, unknown, SchemaType>, res: Response) => {
+    const index = Number(req.params['index']) - 1
+
+    if (Number.isNaN(index)) {
+      return res.render('pages/error', {
+        message: 'Interview not found',
+        status: 404,
+      })
+    }
```
bill-moj, line 47: not sure whether we ever render a POST. i think we always redirect?
`return res.redirect('back')` might work better?

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (!req.journeyData.investigation?.interviews) {
+      req.journeyData.investigation!.interviews = []
+    }
+
+    if (!req.journeyData.investigation?.interviews[index]) {
+      req.journeyData.investigation!.interviews[index] = {}
+    }
+
+    res.redirect(`interview-details/${index + 1}`)
+  }
```
bill-moj, line 68: i think `record-investigation/interview-details` is adding a new interview, while `record-investigation/interview-details/1` means changing the first interview that has been added before.
this is not quite what i expect.

bill-moj: other than the routing/redirection, it's a bit confusing to add an empty interview object..
the current implementation of `/interviews-summary` will misinterpret that as interviews being filled already and show it as completed..
i think it's better to add the interview object when it's submitted.

server/journeys/record-investigation/interview-details/controller.ts
JackReeveMoJ: oh yeah good catch!

JackReeveMoJ: oh yeah, good catch!

server/journeys/record-investigation/interview-details/controller.ts
JackReeveMoJ: Hard to test this one, its more in here for security if someone tries to do this through postman. Originally these were both res.redirect'ing to `pages/404` - I'll update this one to redirect to a 404, feels weird to go `back`, if the user *does* happen to see this for whatever reason, itll be confusing to hit the same page without an error

server/journeys/record-investigation/interview-details/controller.ts
JackReeveMoJ: I've modified the interviews-summary to do the same check where it only lists interviews with an interviewee name recorded

server/journeys/record-investigation/interview-details/tests.cy.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    cy.get('a[href*="interview-details/1#interviewee"]').first().click()
+    verifyDetailsAreRestoredFromJourney()
+
+    cy.url().then(currentUrl => {
+      const modifiedUrl = currentUrl.replace(/interview-details\/(\d+)/g, 'interview-details/$10')
+      cy.visit(modifiedUrl, { failOnStatusCode: false })
+      cy.findByText('404').should('be.visible')
+    })
+  })
```
bill-moj, line 58: probably can simply 
```
cy.visit(`${uuid}/record-investigation/interview-details/10`)
```
?

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (Number.isNaN(index) || (req.journeyData.investigation!.interviews || []).length < index) {
+      return res.status(404).redirect('/pages/404')
+    }
+    
+    if (!req.journeyData.investigation?.interviews) {
+      req.journeyData.investigation!.interviews = []
+    }
+
+    if (!req.journeyData.investigation!.interviews?.[index]) {
```
ReedSoftware, line 44: This line and the ones below it could be changed to:

```suggestion
req.journeyData.investigation!.interviews?.[index] = {
  interviewee: req.body.interviewee
  interviewDate: req.body.interviewDate
  intervieweeRole: req.body.intervieweeRole
  interviewText: req.body.interviewText
}
```

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,54 @@
...
+      req.journeyData.investigation!.interviews![index] = {}
+    }
+
+    req.journeyData.investigation!.interviews![index]!.interviewee = req.body.interviewee
+    req.journeyData.investigation!.interviews![index]!.interviewDate = req.body.interviewDate
```
ReedSoftware, line 49:
```suggestion
```

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,54 @@
...
+    req.journeyData.investigation!.interviews![index]!.intervieweeRole = req.body.intervieweeRole
```
ReedSoftware, line 50:
```suggestion
```

server/journeys/record-investigation/interview-details/controller.ts
```diff
@@ -0,0 +1,54 @@
...
+    req.journeyData.investigation!.interviews![index]!.interviewText = req.body.interviewText
```
ReedSoftware, line 51:
```suggestion
```

server/journeys/record-investigation/interview-details/controller.ts
ReedSoftware: Added this and removing the lines below to a batch suggestion :)

bill-moj (approved):

PR #119, opened 2024-08-12
Remove ccrd lib and related refs

ReedSoftware:


assets/scss/application.scss
```diff
@@ -9,7 +9,7 @@ $govuk-page-width: $moj-page-width;
 
 @import 'node_modules/@ministryofjustice/hmpps-connect-dps-components/dist/assets/footer';
 @import 'node_modules/@ministryofjustice/hmpps-connect-dps-components/dist/assets/header-bar';
-@import 'node_modules/hmpps-court-cases-release-dates-design/hmpps/assets/stylesheets/all';
 
 @import './components/header-bar';
 @import './local';
+@import './server/views/partials/miniProfile/_all.scss'
```
bill-moj, line 15: the new file might go into `/assets/components/_mini-profile.scss` and imported as :
```
@import './components/mini-profile';
```
in the same format as header-bar above?

ReedSoftware: It could, but that would go against keeping all relevant files in a similar place like we've been doing elsewhere.

bill-moj: then can it be `@import './server/views/partials/miniProfile/all'`? i'm not familiar with scss syntax/naming convention

bill-moj (approved):

assets/scss/application.scss
ReedSoftware: Yes it can :) done

bill-moj (approved):

PR #125, opened 2024-08-14
fix path screen/screen -> screen

bill-moj:


server/middleware/redirectCheckAnswersMiddleware.ts
```diff
@@ -5,10 +5,13 @@ export default function redirectCheckAnswersMiddleware(excludePaths: RegExp[] =
 
   router.use((req, res, next) => {
     if (req.originalUrl && !excludePaths.some(itm => req.originalUrl.match(itm))) {
+      const subPaths = req.originalUrl.split('/')
+      const checkAnswersUrl = subPaths.length === 3 ? `${subPaths[2]}/check-answers` : 'check-answers'
```
JackReeveMoJ, line 9: Do we really have to check this? Can't we just check any of subPaths for being equal to `check-answers` ?

JackReeveMoJ (approved):

server/middleware/redirectCheckAnswersMiddleware.ts
bill-moj: yes.
we need this check so `/:journeyId/record-decision` (first question) will redirect to `/:journeyId/record-decision/check-answers`.
and `/:journeyId/record-decision/second-question` will also redirect to `/:journeyId/record-decision/check-answers`.

PR #132, opened 2024-08-16
feat: IaR csip record view after investigation

ReedSoftware:


integration_tests/mockApis/csipApi.ts
```diff
@@ -537,6 +498,85 @@ const stubPostInvestigation = () => {
   })
 }
 
+const referralSubmittedCsip = {
+  recordUuid: '02e5854f-f7b1-4c56-bec8-69e390eb8550',
+  prisonNumber: 'A1111AA',
+  prisonCodeWhenRecorded: 'LEI',
+  createdAt: '2024-07-22T11:21:48',
```
bill-moj, line 505: we might have `screeningSubmittedCsip` `investigationSubmittedCsip` `decisionSubmittedCsip` etc similar to this, since each of them will just be the same object with an extra property populated?

JackReeveMoJ (approved):

server/routes/csip-records/components/investigation.njk
```diff
@@ -0,0 +1,112 @@
...
+          text: referral.createdAt | formatDisplayDate
+        }
+      },
+      {
+        key: {
+          text: 'Staff involved'
+        },
+        value: {
+          text: investigation.staffInvolved
+        }
```
bill-moj, line 24:
```suggestion
        value: {
          html: investigation.staffInvolved | escape | nl2br
        }
```

server/routes/csip-records/components/investigation.njk
```diff
@@ -0,0 +1,112 @@
...
+      },
+      {
+        key: {
+          text: 'Evidence secured'
+        },
+        value: {
+          text: investigation.evidenceSecured
+        }
```
bill-moj, line 32:
```suggestion
        value: {
          html: investigation.evidenceSecured | escape | nl2br
        }
```

server/routes/csip-records/components/investigation.njk
```diff
@@ -0,0 +1,112 @@
...
+      },
+      {
+        key: {
+          text: 'Why this occurred'
+        },
+        value: {
+          text: investigation.occurrenceReason
+        }
```
bill-moj, line 40: ditto for the next few. need to preserve the line breaks.

server/routes/csip-records/components/investigation.njk
```diff
@@ -0,0 +1,112 @@
...
+                html: (interview.interviewText or "Not provided") | escape | nl2br
+            }
+        }]
+    }) }}
+{% endfor %}
+
+{% if not investigation.interviews.length %}
+    <p class="govuk-!-margin-bottom-8">No interview details recorded.</p>
+    <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-8">
+{% endif %}
```
bill-moj, line 112: i think this bit can be deleted

bill-moj (changes requested):

integration_tests/mockApis/csipApi.ts
ReedSoftware: I've just been using the same json and overriding the status as seen further up. Perhaps better to rename this to just `csip`? I dont want to create lots of variables that are just for single use.

bill-moj (approved):

PR #141, opened 2024-08-20
feat: turn referral/involvement jest tests into cypress

ReedSoftware:


server/routes/journeys/record-decision/confirmation/tests.cy.ts
```diff
@@ -50,6 +50,36 @@ context('test /record-decision/confirmation', () => {
         nextSteps: `<script>alert('xss-nextSteps');</script>`,
         actionOther: `<script>alert('xss-actionOther');</script>`,
       },
+      csipRecord: {
+        status: 'PLAN_PENDING',
```
JackReeveMoJ, line 54: The test now expects this to be whatever status is set in the cypress stub. You might be able to get away with removing the whole thing? `cy.task('stubCsipRecordGetSuccess')` should cover this

server/routes/journeys/referral/involvement/tests.cy.ts
```diff
@@ -0,0 +1,218 @@
...
+    )
+    cy.get('p')
+      .contains(/select how the prisoner was involved in the incident/i)
+      .should('be.visible')
+    cy.get('p')
+      .contains(/select if any staff were assaulted during the incident or not/i)
+      .should('be.visible')
+  }
+
+  const setupDataSignInAndGo = () => {
```
JackReeveMoJ, line 149: If you visit the start page for the journey, inject your test data and then manually visit the page under test we shouldn't need to repeat ourselves with all of this csip mock data. See the record-decision/confirmation tests

server/routes/journeys/record-decision/confirmation/tests.cy.ts
ReedSoftware: Good point, done!

server/routes/journeys/referral/involvement/tests.cy.ts
ReedSoftware: @JackReeveMoJ so if I did that, I'd be able to remove `csipRecord` entirely?

JackReeveMoJ: Yep, because it should be covered in the initial start page when that grabs the csip mock

ReedSoftware: That's done now :)

server/routes/journeys/screen/confirmation/tests.cy.ts
```diff
@@ -50,10 +50,33 @@ context('test /screen/confirmation', () => {
...
+          },
+          incidentType: {
+            code: 'A',
+          },
+          isSaferCustodyTeamInformed: 'DO_NOT_KNOW',
+          refererArea: {
+            code: 'A',
+          },
+          referredBy: '',
+        },
```
JackReeveMoJ, line 72: Can get rid of these changes too for similar reasons above, I've got a patch PR open to fix the inconsistency on this page

ReedSoftware: @JackReeveMoJ this one already had the csipRecord status set in it like so:
```js
csipRecord: {
    status: 'PLAN_PENDING',
},```

 - am I ok to just remove the entire csipRecord, or is the test using that status?

bill-moj: you can remove both this and prisoner. they are populated by wiremock stubs.

JackReeveMoJ: PRs just gone in to correct this, I'd say revert all changes to this file and leave as-is

JackReeveMoJ (approved):

PR #152, opened 2024-08-22
Add custom order for Signed Off By Role options

bill-moj:


server/routes/journeys/record-decision/controller.ts
```diff
@@ -17,4 +19,11 @@ export class RecordDecisionController extends BaseJourneyController {
     req.journeyData.decisionAndActions!.signedOffByRole = req.body.signedOffByRole
     res.redirect('record-decision/conclusion')
   }
+
+  private customRadiosOrder = (items: { value: string; text: string | undefined }[]) => [
+    ...items.filter(({ value }) => value === 'CUSTMAN'),
+    ...items.filter(({ value }) => value === 'THEHOFF'),
+    ...items.filter(({ value }) => !['CUSTMAN', 'THEHOFF', 'OTHER'].includes(value)),
+    ...items.filter(({ value }) => value === 'OTHER'),
```
JackReeveMoJ, line 27: This seems like a lot of work just to set a custom order? We're saying here that we want the order to be CUSTMAN, THEHOFF, <anything else>, OTHER

Iterating through this array 4 times seems overkill, I know we won't be having massive arrays so performance isn't really an issue but its a bit awkward to read at first

bill-moj: wrote it like this because to me this format is more readable.
would have written a sorter function instead if performance is a concern.

but then we will let backend handle the ordering so this PR can be deleted.
the custom order logic on the other 3 screens can also be removed later.

bill-moj:
a decision is made to manage ordering on backend api. this PR will be deleted once the issue is resolved.

bill-moj:
ref data to be order on backend, this PR is no longer needed

PR #160, opened 2024-08-27
feat: delete identified need

ReedSoftware:


server/routes/journeys/develop-an-initial-plan/routes.ts
```diff
@@ -16,6 +17,8 @@ function Routes(_services: Services) {
 
   router.use('/identified-needs', IdentifiedNeedsRoutes())
   router.use('/summarise-identified-need', SummariseIdentifiedNeedRoutes())
+  router.use('/delete-identified-need', DeleteIdentifiedNeedRoutes())
```
bill-moj, line 20: do we need this route?

ReedSoftware: It just ensures that if someone types in the url without a `/1` that it gets caught by the GET function and then redirects to the error page

bill-moj: i think it's okay not to handle that and let that go to a 404 page?

ReedSoftware: That's what I'm doing in the GET if there's no index

bill-moj: if user type in a wrong/invalid url, i think it's better not to provide a route to send them to 404 immediately, instead of going into a controller and parse the index and then go to 404?? that way we don't have to be concerned about whether the controller handles undefined/null index params correctly, because the route doesn't even go there.

ReedSoftware: Oh right sorry, we have the error middleware already - ok then :)

bill-moj (approved):

PR #165, opened 2024-08-28
Add placeholder /prisoners/:prisonNumber screen

bill-moj:
image


ReedSoftware:
@bill-moj is there any part of this we'll keep, behaviourally, when we do this "properly"? If so, we should have tests for those parts at least. If not, no worries :)

ReedSoftware (approved):

PR #170, opened 2024-08-29
feat: diap cya

ReedSoftware:


server/routes/journeys/develop-an-initial-plan/routes.ts
```diff
@@ -10,8 +10,9 @@ import { RecordActionsProgressRoutes } from './record-actions-progress/routes'
 import { DeleteIdentifiedNeedRoutes } from './delete-identified-need/routes'
 import { NextReviewDateRoutes } from './next-review-date/routes'
 import { InterventionDetailsRoutes } from './intervention-details/routes'
+import { PlanCheckAnswersRoutes } from './check-answers/routes'
 
-function Routes(_services: Services) {
+function Routes(services: Services) {
   const { router, get, post } = JourneyRouter()
```
bill-moj, line 16: i actually expected the placeholder to be expanded into `Routes({ csipApiService }: Services)`
so the parent route passes around the services as a whole package, but the routes that actually make use of the services can be explicit about which of the services it uses

server/routes/journeys/develop-an-initial-plan/check-answers/controller.ts
```diff
@@ -0,0 +1,61 @@
...
+        reasonForPlan: plan.reasonForPlan!,
+        identifiedNeeds: plan.identifiedNeeds!.map(
+          ({ closedDate, createdDate, identifiedNeed, intervention, progression, responsiblePerson, targetDate }) => ({
+            createdDate,
+            identifiedNeed,
+            intervention,
+            responsiblePerson,
+            targetDate,
+            closedDate: closedDate!,
+            progression: progression!,
```
bill-moj, line 37: `closedDate` is quite definitely `null` when we develop an initial plan. i don't think we have a question screen for it.
`progression` is optional question, so it can be null and can't be `!`'d here.

bill-moj: `createdDate` is to be auto populated with today's date.
i missed setting this value in the PR for `/develop-an-initial-plan/record-actions-progress`.
i think we either add the value over there or here.
probably better to set the date when we push the subJourney identified need into the identifiedNeeds array.

server/routes/journeys/develop-an-initial-plan/check-answers/controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  checkSubmitToAPI = async (req: Request, res: Response, next: NextFunction) => {
+    const plan = req.journeyData.plan!
+    try {
+      await this.csipApiService.createPlan(req, {
+        caseManager: plan.caseManager!,
+        firstCaseReviewDate: plan.firstCaseReviewDate!,
+        reasonForPlan: plan.reasonForPlan!,
+        identifiedNeeds: plan.identifiedNeeds!.map(
+          ({ closedDate, createdDate, identifiedNeed, intervention, progression, responsiblePerson, targetDate }) => ({
```
JackReeveMoJ, line 30: Not a fan of destructuring when there are so many properties inside it. Makes the line long and unruly. I have the following in my branch, also addresses bill's comments below

```
      await this.csipApiService.createInitialPlan(req, {
        caseManager: plan.caseManager!,
        firstCaseReviewDate: plan.firstCaseReviewDate!,
        reasonForPlan: plan.reasonForPlan!,
        identifiedNeeds: plan.identifiedNeeds!.map(identifiedNeed => ({
          identifiedNeed: identifiedNeed.identifiedNeed!,
          responsiblePerson: identifiedNeed.responsiblePerson!,
          intervention: identifiedNeed.intervention!,
          createdDate: todayString(),
          targetDate: identifiedNeed.targetDate!,
          ...(identifiedNeed.progression ? { progression: identifiedNeed.progression! } : {}),
        })),
      })
   ```

Comment (bill-moj) [reply server/routes/journeys/develop-an-initial-plan/check-answers/controller.ts:30]: or maybe 
```
        identifiedNeeds: plan.identifiedNeeds!.map(identifiedNeed => ({
          ...identifiedNeed,
          createdDate: todayString(),
          ...(identifiedNeed.progression ? { progression: identifiedNeed.progression! } : {}),
        })),
```

the type will be wrong tho lol

ReedSoftware: Hi @JackReeveMoJ I've removed closedDate from this list of properties as we didn't want it, so the total line length is now shorter - I think it would end up being more verbose and less easy to read if I didn't destructure with what's there now. Is that ok, or do you still not like the destructuring?

server/routes/journeys/develop-an-initial-plan/check-answers/controller.ts
ReedSoftware: Done on both counts

server/routes/journeys/develop-an-initial-plan/routes.ts
ReedSoftware: Done

server/routes/journeys/develop-an-initial-plan/check-answers/controller.ts
JackReeveMoJ: I still don't like the destructuring tbh but I'm not gonna block the PR for it if you feel its easier to read this way

JackReeveMoJ (approved):

JackReeveMoJ (approved):

PR #175, opened 2024-09-02
feat: better error page

ReedSoftware:


server/errorHandler.ts
```diff
@@ -11,6 +11,15 @@ export default function createErrorHandler(production: boolean) {
       return res.redirect('/sign-out')
     }
 
+    if (
+      production &&
+      ((error.status >= 500 && error.status < 600) || ['TypeError', 'RangeError', 'URIError'].includes(error.name))
```
JackReeveMoJ, line 16: Any reason specifically for these 3 errors? Are we only interested in these three? What would be including that we don't want if we just looked for "Error"?

JackReeveMoJ: Also, I'm personally not a fan of `if` statements spanning multiple lines. I understand the linter has probably done this to keep line length down (though its not broken up the array which is a bit weird). In previous projects I would break this out into using temporary variables to make it a bit more readable but I understand there may be some hesitance to use variables that are only referenced once. What do you think?

```
const is5xx = error.status >= 500 && error.status < 600
const isError = ['TypeError', 'RangeError', 'URIError'].includes(error.name)

if (production && (is5xx || isError)) {
  return res.render('pages/errorServiceProblem', {
    showBreadcrumbs: true,
  })
}
```

JackReeveMoJ (approved):

server/errorHandler.ts
ReedSoftware: It's not on separate lines now that we've made the condition different :)

PR #178, opened 2024-09-03
MIC-498: CSIP Record Page with updated CTA and Summary Cards

JackReeveMoJ:
image

### DIFF
--- a/server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
+++ b/server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
@@ -65,11 +65,8 @@ context('Make a Referral Journey', () => {
       .clear()
       .type(`${today.getDate()}/${today.getMonth() + 1}/${today.getFullYear()}`)
     cy.findByRole('button', { name: /choose date/i }).click()
-    const currentMonthLong = new Intl.DateTimeFormat('en-GB', { month: 'long' }).format(today)
-    cy.findAllByRole('button', { name: new RegExp(`Excluded date.*${currentMonthLong}.*`, 'i'), hidden: true }).should(
-      'have.length',
-      1,
-    )
+    cy.findByRole('button', { name: /Previous month/i }).click()
+    cy.findAllByRole('button', { name: /Excluded date.*/i, hidden: true }).should('have.length.greaterThan', 27)
 
     cy.get('details').invoke('attr', 'open').should('not.exist')
     cy.get('summary').click()
--- a/server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
+++ b/server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
@@ -1,4 +1,5 @@
 import { v4 } from 'uuid'
+import { getDaysInMonth } from 'date-fns'
 import { injectJourneyDataAndReload } from '../../../../../integration_tests/utils/e2eTestUtils'
 
 const uuid = v4()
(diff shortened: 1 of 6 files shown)


server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
```diff
@@ -65,11 +65,8 @@ context('Make a Referral Journey', () => {
       .clear()
       .type(`${today.getDate()}/${today.getMonth() + 1}/${today.getFullYear()}`)
     cy.findByRole('button', { name: /choose date/i }).click()
-    const currentMonthLong = new Intl.DateTimeFormat('en-GB', { month: 'long' }).format(today)
-    cy.findAllByRole('button', { name: new RegExp(`Excluded date.*${currentMonthLong}.*`, 'i'), hidden: true }).should(
-      'have.length',
-      1,
-    )
+    cy.findByRole('button', { name: /Previous month/i }).click()
+    cy.findAllByRole('button', { name: /Excluded date.*/i, hidden: true }).should('have.length.greaterThan', 27)
```
ReedSoftware, line 69: Won't this fail when we're nearer the end of the month?

JackReeveMoJ: It's always going back to the previous month, so they should always be all excluded. It's not a great test for "in the past" but it will at least catch whether its been specified or not

JackReeveMoJ: For some reason I thought it shows days in previous month's like some other date pickers do, glad it doesn't! Updated to check for "today - 1" days in the current month and making sure last month is excluded (for when the test runs on the first day of the month)

server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
```diff
@@ -65,10 +66,11 @@ context('Make a Referral Journey', () => {
...
       .type(`${today.getDate()}/${today.getMonth() + 1}/${today.getFullYear()}`)
     cy.findByRole('button', { name: /choose date/i }).click()
-    const currentMonthLong = new Intl.DateTimeFormat('en-GB', { month: 'long' }).format(today)
-    cy.findAllByRole('button', { name: new RegExp(`Excluded date.*${currentMonthLong}.*`, 'i'), hidden: true }).should(
+    cy.findAllByRole('button', { name: /Excluded date.*/i, hidden: true }).should('have.length', today.getDate() - 1)
+    cy.findByRole('button', { name: /previous month/i }).click()
+    cy.findAllByRole('button', { name: /Excluded date.*/i, hidden: true }).should(
       'have.length',
-      1,
+      getDaysInMonth(new Date(today.getFullYear(), today.getMonth() - 1, 1)),
```
ReedSoftware, line 73: It doesn't actually matter for this test, but getMonth actually returns the month by 0-index, so today.getMonth() - 1 is actually 2 months ago :)

ReedSoftware (approved):

server/routes/journeys/develop-an-initial-plan/next-review-date/tests.cy.ts
JackReeveMoJ: the date constructor also works on 0 indexes so -1 is needed to get the previous month

ReedSoftware: Ah I had misread what you were doing 😅 sorry!

PR #184, opened 2024-09-06
MIC-466: Content amends to Investigation after 2i feedback

JackReeveMoJ:


server/routes/journeys/record-investigation/why-behaviour-occurred/view.njk
```diff
@@ -13,24 +13,17 @@
     </label>
     <form method="post">
         <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
-        {{ govukDetails({
-            summaryText: "What type of information to include",
```
bill-moj, line 17: `{% from "govuk/components/details/macro.njk" import govukDetails %}` is unused and can be deleted

bill-moj (changes requested):
missing amendment on /record-investigation screen, `cannot *save* yet`

bill-moj (approved):

PR #194, opened 2024-09-11
MIC-524: Cypress/Jest code coverage report

JackReeveMoJ:


.circleci/config.yml
```diff
@@ -100,14 +104,15 @@ jobs:
           name: Wait for node app to start
           command: sleep 5
       - run:
-          name: integration tests
-          command: npm run int-test
+          name: Integration tests
+          command: circleci tests glob "**/*.cy.ts" | circleci tests run --command="xargs npx cypress run --reporter cypress-circleci-reporter --spec" --verbose --split-by=timings
```
ReedSoftware, line 108: How come we're using this special command now instead of `npm run int-test`?

server/app.ts
```diff
@@ -3,6 +3,8 @@ import express from 'express'
 import dpsComponents from '@ministryofjustice/hmpps-connect-dps-components'
 
 import * as Sentry from '@sentry/node'
+// @ts-expect-error Import untyped middleware for cypress coverage
```
ReedSoftware, line 6: There is a DT implementation apparently https://www.npmjs.com/package/@types/cypress__code-coverage - worth a try?

.circleci/config.yml
JackReeveMoJ: This is for CircleCI to split the tests up across its different instances. DIdn't want to include this in the package.json as its specific to CircleCI

server/app.ts
JackReeveMoJ: We do have this in the package.json, but it still complains

.circleci/config.yml
ReedSoftware: Ah right and there's no way to encode this into the config instead and still run this normally via `npm run int-test`? If not np

server/app.ts
ReedSoftware: What a pain. Ok, no worries then.

ReedSoftware (approved):

.circleci/config.yml
JackReeveMoJ: Not that I'm aware of, I believe this code needs to run within the context of the pipeline and we let circleCI do the splitting

PR #201, opened 2024-09-12
MIC-503 /update-referral/details screen

bill-moj:
MIC-503

image

image


server/utils/utils.ts
```diff
@@ -51,9 +51,9 @@ export const sentenceCase = (val: string, startsWithUppercase: boolean = true):
   return startsWithUppercase ? sentence.charAt(0).toUpperCase() + sentence.slice(1) : sentence
 }
 
-export const getNonUndefinedProp = <T>(referral: T, key: keyof T) => {
+export const getNonUndefinedProp = <T>(referral: T, key: keyof T, mapper?: (obj: unknown) => string | null) => {
```
JackReeveMoJ, line 54: minor: the first param shouldn't be called referral here as it exists outside of the context of referrals (ie, its a generic util method that could/will get used in future flows)

JackReeveMoJ (approved):

ReedSoftware (approved):

PR #224, opened 2024-09-25
feat: update why behaviour occurred

ReedSoftware:


server/routes/journeys/update-investigation/why-behaviour-occurred/controller.ts
```diff
@@ -0,0 +1,36 @@
+import { NextFunction, Request, Response } from 'express'
+import { SchemaType } from '../../record-investigation/why-behaviour-occurred/schemas'
+import { MESSAGE_INVESTIGATION_UPDATED, PatchInvestigationController } from '../../base/patchInvestigationController'
+import { generateSaveTimestamp, getMaxCharsAndThresholdForAppend } from '../../../../utils/appendFieldUtils'
+
+export class UpdateWhyBehaviourOccurredController extends PatchInvestigationController {
+  GET = async (req: Request, res: Response) => {
+    req.journeyData.isUpdate = true
```
bill-moj, line 8: this has been set in /update-investigation/controller so we don't have to set it again here.
i think i also forgot to take this out for /update-investigation/staff-involved/controller

ReedSoftware: Ah thanks - removed it from there too.

bill-moj (approved):

PR #233, opened 2024-09-26
MIC-576: UaI: Evidence secured

JackReeveMoJ:


server/routes/journeys/update-investigation/evidence-secured/controller.ts
```diff
@@ -0,0 +1,37 @@
...
+  GET = async (req: Request, res: Response) => {
+    req.journeyData.isUpdate = true
+    const currentEvidenceSecured = req.journeyData.csipRecord!.referral.investigation!.evidenceSecured
+    res.render('record-investigation/evidence-secured/view', {
+      currentEvidenceSecured,
+      evidenceSecured: res.locals.formResponses?.['evidenceSecured'],
+      isUpdate: true,
+      recordUuid: req.journeyData.csipRecord!.recordUuid,
+      ...getMaxCharsAndThresholdForAppend(res.locals.user.displayName, currentEvidenceSecured),
+    })
```
bill-moj, line 16: `req.journeyData.isUpdate = true` is already set by `/update-investigation/controller` and isn't needed here.

we need a back link on this screen according to the ticket a/c

server/routes/journeys/record-investigation/evidence-secured/schemas.ts
```diff
@@ -1,14 +1,28 @@
 import z from 'zod'
+import { Request, Response } from 'express'
 import { createSchema } from '../../../../middleware/validationMiddleware'
+import { getMaxCharsAndThresholdForAppend } from '../../../../utils/appendFieldUtils'
 
-const REASONS_MSG = 'Enter a description of the evidence secured'
-const TOO_LONG_ERROR_MSG = 'Description of the evidence secured must be 4,000 characters or less'
+export const schemaFactory = async (req: Request, res: Response) => {
+  const ERROR_MSG = 'Enter a description of the evidence secured'
+  const UPDATE_ERROR_MSG = 'Enter an update to the description of the evidence secured'
```
bill-moj, line 8: these can go outside of the schemaFactory's scope.

server/routes/journeys/record-investigation/evidence-secured/view.njk
```diff
@@ -1,30 +1,36 @@
...
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
 
 {% extends "partials/layout.njk" %}
 
-{% set pageTitle = "Evidence secured - Record a CSIP investigation" %}
+{% set pageTitle = "Evidence secured - " + ('Update' if isUpdate else 'Record') + " a CSIP investigation" %}
 
 {% block innerContent %}
-    <span class="govuk-caption-l">Record a CSIP investigation</span>
+    <span class="govuk-caption-l">{{'Update' if isUpdate else 'Record'}} a CSIP investigation</span>
```
bill-moj, line 11: there's a `{% include 'base/components/investigation-caption.njk' %}` for this caption

server/routes/journeys/record-investigation/evidence-secured/view.njk
```diff
@@ -1,30 +1,36 @@
 {% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
 {% from "govuk/components/details/macro.njk" import govukDetails %}
-{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "partials/submitButton/macro.njk" import submitButton %}
```
bill-moj, line 3: we don't have to import `govukDetails` for this view?

server/routes/journeys/routes.ts
```diff
@@ -24,14 +24,27 @@ export const JourneyRoutes = (services: Services) => {
...
+          destination[key] = {}
+        }
+        mergeObjects(destination[key] as Record<string, unknown>, value)
+      } else {
+        // @ts-expect-error unexpected types
+        destination[key] = value
+      }
+    })
+  }
+
```
bill-moj, line 42: looks like it belongs to an util class.
should we import some library such as lodash to handle deep merge?

ReedSoftware: I agree with Bill in that this should either go in a util class or just be removed as a function and move the code inside the `if (process.env.NODE_ENV === 'e2e-test') {`

server/routes/journeys/record-investigation/evidence-secured/view.njk
JackReeveMoJ: you're right, we don't. looks like this was missed 2 months ago

server/routes/journeys/routes.ts
ReedSoftware: I actually prefer just moving the code inside the `if`. If we are having to do a deep merge in more than one place, I think I would prefer to use a library for it.

JackReeveMoJ: Moved the declaration to be inside the `if`. it is a recursive function though so it needs to stay as a function

ReedSoftware: Ah of course :)

bill-moj (approved):

PR #234, opened 2024-09-26
Mic 580 update protective factors

ReedSoftware:


JackReeveMoJ (commented):
We're missing tests here for char remaining count showing chars left on page load when theres already 3000+

server/routes/journeys/record-investigation/protective-factors/view.njk
```diff
@@ -1,41 +1,50 @@
...
+                <p>Protective factors are anything that helps a prisoner deal more effectively
+                with stressful events or triggers. </p>
+                <p>Include information on any: </p>
+                <ul>
+                <li>positive influences on their behaviour, both historic and recent</li>
+                <li>skills, strengths, resources, supports or coping strategies</li>
+                </ul>'
+            }) }}
+        {% endif %}
+        {% if isUpdate and currentProtectiveFactors %}
```
JackReeveMoJ, line 31: don't need to do both checks here, just checking for the presence of `currentProtectiveFactors` should be enough

ReedSoftware: We do as we shouldnt show the inset text if currentProtectiveFactors is falsey

JackReeveMoJ: Yeah that's fine, a falsey value for `currentProtectiveFactors` (`undefined`, `null`, or `''`) will fail the if check

ReedSoftware: Done

ReedSoftware:
> We're missing tests here for char remaining count showing chars left on page load when theres already 3000+

Done

integration_tests/mockApis/csipApi.ts
```diff
@@ -573,6 +573,50 @@ const stubCsipRecordSuccessAwaitingDecisionNoInterviews = () => {
...
+            personsUsualBehaviour: 'a great person',
+            personsTrigger: 'spiders',
+            protectiveFactors: 'a'.repeat(3000),
+          },
+        },
+      },
+    },
+  })
+}
+
```
JackReeveMoJ, line 619: This is why I've modified the inject route to be able to merge objects - so that we don't need to create a whole new object like this just to change one property. 

Do you want to wait for my PR to go in first or should we make a remedial PR to change this and any other examples of this after?

ReedSoftware: Let's make a remedial PR - I know there's more than one instance of this.

JackReeveMoJ (approved):

PR #243, opened 2024-09-30
MIC-582 /update-decision screen

bill-moj:
MIC-582

image

image

### DIFF
--- a/server/routes/csip-records/controller.ts
+++ b/server/routes/csip-records/controller.ts
@@ -4,7 +4,6 @@ import CsipApiService from '../../services/csipApi/csipApiService'
 import PrisonerSearchService from '../../services/prisonerSearch/prisonerSearchService'
 import { FLASH_KEY__CSIP_SUCCESS_MESSAGE } from '../../utils/constants'
 import { interviewSorter } from '../../utils/sorters'
-import logger from '../../../logger'
 
 export class CsipRecordController {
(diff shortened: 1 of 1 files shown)


server/routes/csip-records/controller.ts
```diff
@@ -19,6 +15,7 @@ export class CsipRecordController {
   private getRecordInfo = async (req: Request) => {
     const { recordUuid } = req.params
     const record = await this.csipApiService.getCsipRecord(req, recordUuid!)
+    logger.info(record)
```
ReedSoftware, line 18: Did you mean to leave this in?

server/routes/csip-records/view.njk
```diff
@@ -88,7 +88,9 @@
     {% if decision %}
       {% include './components/decision.njk' %}
     {% endif %}
-    {% include './components/investigation.njk' %}
+    {% if not isUpdate or not decision %}
```
ReedSoftware, line 91: Wont this try to show the investigation template if we're in update mode, even if an investigation doesn't exist?

bill-moj: update-xxx controller redirects to csip-records page if it's not in the right status, so it wouldn't reach this bit.
we shouldn't set `referralTabSelected = false` if there isn't an investigation tab?

PR #248, opened 2024-10-02
MIC-619: Update a Plan - Case Management

JackReeveMoJ:
image

image

### DIFF
--- a/server/routes/journeys/update-plan/case-management/controller.ts
+++ b/server/routes/journeys/update-plan/case-management/controller.ts
@@ -20,7 +20,6 @@ export class CaseManagementController extends PatchPlanController {
         reasonForPlan: req.body.reasonForPlan,
         caseManager: req.body.caseManager,
       },
-      successMessage: 'You’ve updated the case management information.',
     })
 
   POST = async (req: Request<unknown, unknown, SchemaType>, res: Response) => {
--- a/server/routes/journeys/update-plan/case-management/controller.ts
+++ b/server/routes/journeys/update-plan/case-management/controller.ts
@@ -1,8 +1,8 @@
 import { NextFunction, Request, Response } from 'express'
 import { SchemaType } from './schemas'
-import { UpdatePlanController } from '../controller'
+import { PatchPlanController } from '../../base/patchPlanController'
 
-export class CaseManagementController extends UpdatePlanController {
+export class CaseManagementController extends PatchPlanController {
   GET = async (req: Request, res: Response) => {
     res.render('update-plan/case-management/view', {
       reasonForPlan: res.locals.formResponses?.['reasonForPlan'] ?? req.journeyData.plan?.reasonForPlan,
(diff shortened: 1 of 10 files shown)


server/routes/journeys/update-plan/case-management/controller.ts
```diff
@@ -0,0 +1,29 @@
+import { NextFunction, Request, Response } from 'express'
+import { SchemaType } from './schemas'
+import { UpdatePlanController } from '../controller'
+
+export class CaseManagementController extends UpdatePlanController {
+  GET = async (req: Request, res: Response) => {
+    res.render('update-plan/case-management/view', {
+      reasonForPlan: res.locals.formResponses?.['reasonForPlan'] || req.journeyData.plan?.reasonForPlan,
+      caseManager: res.locals.formResponses?.['caseManager'] || req.journeyData.plan?.caseManager,
```
bill-moj, line 9: `??` instead of `||`

server/routes/journeys/update-plan/case-management/controller.ts
```diff
@@ -0,0 +1,29 @@
...
+  checkSubmitToAPI = async (req: Request<unknown, unknown, SchemaType>, _: Response, next: NextFunction) =>
+    this.submitChanges({
+      req,
+      next,
+      changes: {
+        reasonForPlan: req.body.reasonForPlan,
+        caseManager: req.body.caseManager,
+      },
+      successMessage: 'You’ve updated the case management information.',
+    })
```
bill-moj, line 24: UpdatePlanController.submitChanges has the message content covered, you don't need to pass in `successMessage`

server/routes/journeys/update-plan/controller.ts
```diff
@@ -26,7 +57,7 @@ export class UpdatePlanController extends BaseJourneyController {
       ...getNonUndefinedProp(plan, 'caseManager'),
       ...getNonUndefinedProp(plan, 'reasonForPlan'),
       ...getNonUndefinedProp(plan, 'firstCaseReviewDate'),
-      identifiedNeeds: plan.identifiedNeeds.sort(identifiedNeedSorter).map(need => ({
+      identifiedNeeds: (plan.identifiedNeeds || []).sort(identifiedNeedSorter).map(need => ({
```
bill-moj, line 60: can `plan.identifiedNeeds` be nullish?
i thought backend would at least return an empty array? (not sure)

JackReeveMoJ: Yeah it was crashing on one of the csip records I was trying earlier

server/routes/journeys/update-plan/case-management/controller.ts
JackReeveMoJ: No, it doesn't. We pass it through into submitChanges because the text can change depending on what we're updating

bill-moj (approved):

PR #279, opened 2024-10-17
MIC-673: RaR: Close CSIP

JackReeveMoJ:
image

image

image

### DIFF
--- a/server/routes/journeys/record-review/close-csip/tests.cy.ts
+++ b/server/routes/journeys/record-review/close-csip/tests.cy.ts
@@ -25,7 +25,9 @@ context('test /record-review/close-csip', () => {
     cy.visit(START_URL, { failOnStatusCode: false })
     cy.visit(PAGE_URL)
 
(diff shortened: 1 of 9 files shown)


server/routes/journeys/record-review/close-csip/tests.cy.ts
```diff
@@ -0,0 +1,99 @@
...
+    cy.task('stubCsipRecordGetSuccess')
+  })
+
+  it('should try out all cases - with no open identified needs', () => {
+    cy.signIn()
+    cy.visit(START_URL, { failOnStatusCode: false })
+    cy.visit(PAGE_URL)
+
+    cy.get('.govuk-grid-column-two-thirds').children().should('have.length', 5)
+
```
bill-moj, line 29: it's not obvious what is being tested here.
can it instead findBy something else and `should('not.exists')` so we know what you're trying to assert?

bill-moj (approved):

PR #283, opened 2024-10-18
MIC-684 Update a Review - initial screen

bill-moj:
MIC-684

image

image

image


server/routes/journeys/record-review/check-answers/view.njk
```diff
@@ -118,24 +118,21 @@
                     text: 'Contribution'
                 },
                 value: {
-                    text: (attendee.contribution or 'Not provided') | escape | boldAppendStamp | nl2br
+                    html: (attendee.contribution or 'Not provided') | escape | boldAppendStamp | nl2br
```
iterative-it, line 121: These will follow a direct text edit model rather than an append model won't they? I guess no harm having the boldAppendStamp call in this in case the model gets switched

server/routes/journeys/update-review/controller.ts
```diff
@@ -0,0 +1,53 @@
...
+      (prev, cur) => (prev!.reviewSequence > cur.reviewSequence ? prev : cur),
+      record.plan.reviews[0],
+    )!
+
+    req.journeyData.review = {
+      reviewUuid: review.reviewUuid,
+      ...getNonUndefinedProp(review, 'reviewDate'),
+      ...getNonUndefinedProp(review, 'nextReviewDate'),
+      ...getNonUndefinedProp(review, 'summary'),
+      outcome: review.csipClosedDate ? 'CLOSE_CSIP' : 'REMAIN_ON_CSIP',
```
iterative-it, line 33: This will "fix" the bad NOMIS data removing any other actions they'd set as well as the ones where both remain and close are set won't it? That seems fine to me especially as it's audited. Have we checked with the team that it's what was intended?

server/routes/journeys/record-review/check-answers/view.njk
bill-moj: it's the Record a Review CYA screen, so it wouldn't have been appended anyway and didn't need that filter.
we seem to have copy and pasted it on all the 4k char fields tho.

server/utils/sorters.ts
```diff
@@ -14,3 +14,8 @@ export const identifiedNeedSorter = (
   a.createdDate.localeCompare(b.createdDate) || // 2. sort by ascending Create Date, if tie then
   a.identifiedNeed.localeCompare(b.identifiedNeed) || // 3. sort by title alphabetically, if tie then
   a.identifiedNeedUuid.localeCompare(b.identifiedNeedUuid) // 4. sort by UUID alphabetically
+
+export const attendeeSorter = (a: components['schemas']['Attendee'], b: components['schemas']['Attendee']) =>
+  Number(!a.name) - Number(!b.name) || // 1. sort nullish name to bottom, if tie then
+  (a.name && b.name && a.name.localeCompare(b.name)) || // 2. sort by non-null name alphabetically, if tie then
+  a.attendeeUuid.localeCompare(b.attendeeUuid) // 3. sort by UUID alphabetically
```
iterative-it, line 21: Final tie break by created date would be sufficient here I think

server/routes/journeys/update-review/controller.ts
bill-moj: yes. cfm'd with Theo it's okay to use csipClosedDate.
the alternative logic i suggested was to check whether `actions` contain `CLOSE_CSIP`, but checking csipCLosedDate is more straightforward. and Anthony has confirmed that all prod reviews with `CLOSE_CSIP` also have a closedDate.

server/utils/sorters.ts
bill-moj: multiple attendees on the same review would be created at the same time, i assume they might have the exact same timestamp? that's why i think tiebreak with uuid would give the best certainty

iterative-it (approved):
LGTM

PR #293, opened 2024-10-24
MIC-688: UaR: Close CSIP

JackReeveMoJ:
image

### DIFF
--- a/server/routes/journeys/update-review/close-csip/controller.ts
+++ b/server/routes/journeys/update-review/close-csip/controller.ts
@@ -4,7 +4,6 @@ import { PatchReviewController } from '../../base/patchReviewController'
 export class UpdateCloseCsipController extends PatchReviewController {
   GET = async (req: Request, res: Response) => {
     res.render('record-review/close-csip/view', {
-      closeCsip: res.locals.formResponses?.['closeCsip'] ?? req.journeyData.review?.outcome,
       isUpdate: true,
       backUrl: `outcome`,
       recordUuid: req.journeyData.csipRecord!.recordUuid,
--- a/server/routes/journeys/update-review/close-csip/controller.ts
+++ b/server/routes/journeys/update-review/close-csip/controller.ts
@@ -3,7 +3,6 @@ import { PatchReviewController } from '../../base/patchReviewController'
 
 export class UpdateCloseCsipController extends PatchReviewController {
   GET = async (req: Request, res: Response) => {
-    req.journeyData.isUpdate = true
     res.render('record-review/close-csip/view', {
       closeCsip: res.locals.formResponses?.['closeCsip'] ?? req.journeyData.review?.outcome,
       isUpdate: true,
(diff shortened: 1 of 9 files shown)


server/routes/journeys/update-review/close-csip/controller.ts
```diff
@@ -0,0 +1,29 @@
+import { NextFunction, Request, Response } from 'express'
+import { PatchReviewController } from '../../base/patchReviewController'
+
+export class UpdateCloseCsipController extends PatchReviewController {
+  GET = async (req: Request, res: Response) => {
+    req.journeyData.isUpdate = true
```
bill-moj, line 6: this is already set at /update-review/controller, we shouldn't need this here

server/routes/journeys/update-review/close-csip/controller.ts
```diff
@@ -0,0 +1,29 @@
...
+    res.render('record-review/close-csip/view', {
+      closeCsip: res.locals.formResponses?.['closeCsip'] ?? req.journeyData.review?.outcome,
```
bill-moj, line 8: i don't think this is needed?

server/routes/journeys/update-review/close-csip/controller.ts
```diff
@@ -0,0 +1,29 @@
...
+      isUpdate: true,
+      backUrl: `outcome`,
+      recordUuid: req.journeyData.csipRecord!.recordUuid,
+      openIdentifiedNeeds: (req.journeyData.csipRecord?.plan?.identifiedNeeds || []).filter(o => !o.closedDate).length,
```
bill-moj, line 12: probably not needed

server/routes/journeys/update-review/next-review-date/controller.ts
```diff
@@ -21,6 +21,7 @@ export class UpdateNextReviewDateController extends PatchReviewController {
       changes: {
         nextReviewDate: req.body.nextReviewDate,
       },
+      message: 'You’ve updated the next review date',
```
bill-moj, line 24: thanks! i missed this a/c!

server/routes/journeys/update-review/close-csip/controller.ts
JackReeveMoJ: It's needed to display the warning text of "X open identified needs"

server/routes/journeys/update-review/close-csip/controller.ts
JackReeveMoJ: It might be needed on the small chance we get an API error and they try again

bill-moj: it's in the session. if they lost it, then they would have lost req.journeyData.csipRecord as well, and need to restart the journey from the beginning

JackReeveMoJ: `req.journeyData.review?.outcome` would be "open" on a page refresh tho because the POST handler hasnt hit if there was an API error. So we bring the last entered value back from `formResponses`. Am I missing something?

bill-moj (approved):

PR #296, opened 2024-10-25
MIC-693 /manage-csips screen

bill-moj:
MIC-693


server/routes/manage-csips/controller.ts
```diff
@@ -0,0 +1,107 @@
...
+        const [sortingKey, sortingDirection] = (sort as string).split(',')
+        if (
+          ['name', 'location', 'referralDate', 'caseManager', 'nextReviewDate', 'status'].includes(sortingKey ?? '') &&
+          ['asc', 'desc'].includes(sortingDirection ?? '')
+        ) {
+          req.session.searchCsipParams.sort = sort as string
+        }
+      }
+
+      return res.redirect('manage-csips')
```
ReedSoftware, line 72: Why do we need to redirect if we have any of the search/query/sort? Can we not just continue with code execution using the values you've just put into searchCsipParams?

server/routes/manage-csips/view.njk
```diff
@@ -0,0 +1,168 @@
...
+            text: 'Support through ACCT'
+          }
+        ] | setSelected(status or '')
+      }) }}
+
+      {{ govukButton({
+        text: "Apply filters",
+        preventDoubleClick: true
+      }) }}
+      <a class="govuk-link govuk-link--no-visited-state govuk-body no-label-form-clear-link" href="?clear=true">Clear</a>
```
ReedSoftware, line 88: Dont think you need `=true` - the presence of the query param should be enough?

server/utils/viewUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    endIndex = match.index + match[0].length
+  })
+
+  res.push(val.substring(endIndex))
+
+  return res.join('')
+}
+
+export const shyHyphens = (text: string | undefined, maxLength: number) => {
```
ReedSoftware, line 63: Maybe call this `hyphenateString`?

ReedSoftware (approved):
Some minor comments - nothing to block this

assets/scss/components/_horizontal-search-form.scss
```diff
@@ -0,0 +1,71 @@
...
+
+    > a {
+      margin-right: govuk-spacing(3);
+    }
+  }
+}
+
+.no-label-form-clear-link {
+  margin-bottom: 10px;
+  margin-top: 0 !important;
```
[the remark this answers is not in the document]
bill-moj: i believe so. it's copypasta from prisoner profile tho i didn't write this lol

server/routes/manage-csips/controller.ts
bill-moj: i want to eliminate potential confusion whether the filter and sort are governed by session or by url query string, so i consume the query strings and populate them into session. 

switching to use url query is technically feasible, but that would mean we need to convert post form data of search string and status filter into query string, and concatenating the query strings for all those page buttons and sort buttons will be a bit horrible.

server/routes/manage-csips/view.njk
bill-moj: clear=any_value would work. just having ?clear or empty ?clear= wouldn't work because i'm only checking `if (clear)` not `if (clear !== undefined)`

PR #301, opened 2024-10-29
Mic 608 active prisons flag

ReedSoftware:


server/middleware/checkPopulateUserCaseloads.ts
```diff
@@ -0,0 +1,63 @@
...
+      !req.url.includes('service-not-enabled')
+    ) {
+      try {
+        if (res.locals.feComponentsMeta?.caseLoads) {
+          res.locals.user.caseloads = res.locals.feComponentsMeta.caseLoads
+        }
+        if (res.locals.feComponentsMeta?.activeCaseLoad) {
+          res.locals.user.activeCaseLoadId = res.locals.feComponentsMeta.activeCaseLoad.caseLoadId
+        }
+        const refetchCaseloads = !res.locals.user.caseloads || res.locals.user.caseloads.length > 1
```
iterative-it, line 35: This will refetch it even if it's come from the fe components if it has > 1 case load. Only needs the first check I think

ReedSoftware: Done

server/@types/express/index.d.ts
```diff
@@ -3,13 +3,15 @@ import { Breadcrumbs } from '../../middleware/breadcrumbs'
...
 import Prisoner from '../../services/prisonerSearch/prisoner'
+import { CaseLoad } from '../../interfaces/caseLoad'
 
 export declare module 'express-session' {
   // Declare that the session will potentially contain these additional fields
   interface SessionData {
     returnTo: string
     nowInMinutes: number
     journeyDataMap: JourneyDataMap
+    userCaseloads?: CaseLoad[]
```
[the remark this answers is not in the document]
ReedSoftware: Done

server/middleware/checkPopulateUserCaseloads.ts
```diff
@@ -0,0 +1,63 @@
...
+  const router = Router()
+
+  router.get('/service-not-enabled', (_req, res) => {
+    res.status(200)
+    res.render('service-not-enabled')
+  })
+
+  router.use(async (req, res, next) => {
+    const splitUrl = req.url.split('/').filter(Boolean)
+    if (
```
iterative-it, line 22: Missed this last review. The population of the case load properties is separate from the check for service switch on. The population will need to be done outside of the if so that they are available for App Insights and other uses

iterative-it (dismissed):
LGTM

iterative-it (approved):
LGTM

PR #310, opened 2024-11-01
MIC-742: Update a CSIP referral amendments

JackReeveMoJ:


server/middleware/validationMiddleware.ts
```diff
@@ -86,7 +86,8 @@ export const validate = (schema: z.ZodTypeAny | SchemaFactory): RequestHandler =
       )
     }
     req.flash(FLASH_KEY__VALIDATION_ERRORS, JSON.stringify(deduplicatedFieldErrors))
-    return res.redirect('back')
+    // Remove any hash from the URL by appending an empty hash string
+    return res.redirect(`${req.baseUrl}#`)
```
bill-moj, line 90: there are 7 more places where we used `res.redirect('back')`
can we reuse the same function here?

assets/scss/local.scss
```diff
@@ -34,4 +34,9 @@
 
 .min-width-one-quarter {
   min-width: 350px;
+}
+
+.dropshadow {
+  border: 1px solid govuk-colour('dark-grey');
+  box-shadow: 0 5px govuk-colour('mid-grey');
```
bill-moj, line 41: i don't see an a/c in the ticket about box shadow. where does this come from?

server/middleware/validationMiddleware.ts
JackReeveMoJ: This comes from the A/C to highlight/focus the error box when there's been a validation error. If the url ends with an anchor? (#) to a control then that control gets focus instead. `res.redirect('back')` keeps the old # reference and this was the only way I could figure around removing it

bill-moj: the comment doesn't seem to mean that tho lol

bill-moj (approved):

PR #312, opened 2024-11-04
chore: content amendments - RaI and UaI

ReedSoftware:


server/routes/journeys/record-investigation/confirmation/view.njk
```diff
@@ -2,7 +2,7 @@
 
 {% extends "partials/layout.njk" %}
 
-{% set pageTitle = "Confirmation - Record a CSIP investigation" %}
+{% set pageTitle = "CSIP investigation recorded" %}
```
bill-moj, line 6: conflict with MIC-734 but i think MIC-744 is correct

bill-moj (approved):

PR #323, opened 2024-11-07
MIC-642: End to end tests for record investigation

JackReeveMoJ:


server/routes/journeys/record-investigation/e2e.cy.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    cy.url().should('to.match', /confirmation$/)
+
+    cy.findByRole('link', { name: 'View all CSIPs for Leeds (HMP)' })
+      .should('be.visible')
+      .and('have.attr', 'href')
+      .and('match', /\/manage-csips\?clear=true$/)
+
+    cy.go('back')
+    cy.url().should('to.match', /confirmation$/)
```
bill-moj, line 123: as James explained in his PR, it's better to wait a sec before verifying that the redirection logic works.

JackReeveMoJ: No need to wait, looks like cypress already waits for a redirection to finish before continuing

bill-moj (approved):

PR #334, opened 2024-11-14
feat: rar e2e

ReedSoftware:


server/routes/journeys/record-review/e2e.cy.ts
```diff
@@ -0,0 +1,132 @@
...
+    cy.contains('dt', 'Attended in person').next().should('include.text', 'Yes')
+    cy.contains('dt', 'Contribution').next().should('include.text', 'a contrib')
+
+    verifyCyaChangeLinks()
+  }
+
+  const verifyCyaChangeLinks = () => {
+    cy.findByRole('link', { name: /Change the details of the review/ }).click()
+    cy.url().should('to.match', /details#summary$/)
+    cy.get('textarea').clear().type('changed summary', { delay: 0 })
```
JackReeveMoJ, line 86: Is it worth using `cy.focused()` for these just as a sanity check that the input element is being focused?

ReedSoftware: @JackReeveMoJ Why do we care about the text area being focused when we type? Im confused 😅

bill-moj: cya page a/c says something about focusing on the specific html input element.
in the unit-esque cypress test for just the check-answers page, we didn't test its interactions with the other pages. 
so i guess that would belong to the full e2e test here if we're concerned about that.

ReedSoftware: Oh you mean when we click the change link - that sort of detail really should belong in the check-answers tests, not this one. Added there :)

JackReeveMoJ (approved):

PR #342, opened 2024-11-22
chore: get audit events working for page view events

ReedSoftware:


server/app.ts
```diff
@@ -53,6 +54,41 @@ export default function createApp(services: Services): express.Application {
...
+        req.query,
+        res.locals.auditEvent,
+        `ACCESS_ATTEMPT_`,
+      )
+    })
+
+    const resRender = res.render
+    res.render = (view: string, options?) => {
+      type resRenderCb = (view: string, options?: object, callback?: (err: Error, html: string) => void) => void
+      ;(resRender as resRenderCb).call(res, view, options, async (err: Error, html: string) => {
```
JackReeveMoJ, line 80: Not a fan of starting a line with `;` usually its considered a code smell in standard.
Would it work declaring something like `const resRender = res.render as (view: string, options?: object, callback?: (err: Error, html: string) => void) => void` ? Might have to use `as any` or `as unknown` first if not

JackReeveMoJ (approved):

PR #348, opened 2024-12-03
feat: add notification if prisoner on csip already

ReedSoftware:


integration_tests/mockApis/auth.ts
```diff
@@ -178,11 +179,12 @@ export default {
   stubAuthManageDetails: manageDetails,
   stubSignIn: (
     userToken: UserToken = { roles: [AuthorisedRoles.ROLE_CSIP_PROCESSOR] },
-  ): Promise<[Response, Response, Response, Response, Response, Response, Response]> =>
+  ): Promise<[Response, Response, Response, Response, Response, Response, Response, Response]> =>
     Promise.all([
       favicon(),
       redirect(),
       signOut(),
+      stubCurrentCsipStatusNoCsip(),
```
bill-moj, line 187: can we add a boolean param to stubSignIn to determine whether to stubCurrentCsipStatusNoCsip or stubCurrentCsipStatusOnCsip?

then we don't have to use two conflict stubs and rely on one to override the other

PR #357, opened 2024-12-06
Refactor journeyData

bill-moj:


server/routes/journeys/base/controller.ts
```diff
@@ -41,4 +41,11 @@ export class BaseJourneyController {
       text: refData.description,
       checked: Array.isArray(value) ? value.includes(refData.code) : refData.code === value,
     }))
+
+  deleteJourneyDataAndGoBackToCsipRecordPage = async (req: Request, res: Response) => {
+    const { recordUuid } = req.journeyData.csipRecord!
+    // @ts-expect-error delete non-optional req.journeyData to free up redis memory
+    delete req.journeyData
```
JackReeveMoJ, line 48: Does this clear in redis as well or just locally?

bill-moj: persistence of req.journeyData is managed by the setupJourneyDataMiddleware
it deletes the data from redis if req.journeyData is undefined (deleted)

PR #394, opened 2025-02-11
MIC-949: Disable Slack notifications

sameer-moj:
Disable Slack notifications for pre-prod and prod deployments


iterative-it (approved):
LGTM

PR #421, opened 2025-03-18
MIAD-357: Fix double tabbing on column headers

JackReeveMoJ:
Before: 

MIAD-357-Before

After:

MIAD-357-After


iterative-it (approved):
LGTM

PR #439, opened 2025-04-14
MIAD-536: Set threshold to 75 where previous thresholds were 100 or undefined

JackReeveMoJ:


sameer-moj (approved):

server/routes/journeys/update-plan/summarise-identified-need/tests.cy.ts
```diff
@@ -58,11 +58,6 @@ context('test /update-plan/summarise-identified-need', () => {
       .and('match', /..\/identified-needs$/)
 
     getSummary().should('be.visible')
-
-    getSummary().clear().type('a'.repeat(999), { delay: 0 })
```
iterative-it, line 62: Should these checks be replaced by any others or are they no longer valid?

JackReeveMoJ: Good catch, there should be a call in this file using the new cypress command but I missed it somehow

iterative-it (approved):
LGTM

PR #460, opened 2025-05-16
MIAD-601 Revamp print button function

bill-moj:
MIAD-601
image

### DIFF
--- a/server/routes/csip-records/view.njk
+++ b/server/routes/csip-records/view.njk
@@ -40,7 +40,7 @@
     }
 
     const hideFromPrint = (...ids) => {
-      for (const block in id) {
+      for (const block in ids) {
         togglePrintBlock(block, false)
       }
     }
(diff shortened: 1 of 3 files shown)


server/routes/csip-records/view.njk
```diff
@@ -1,74 +1,130 @@
...
-        printButton.classList.remove('govuk-!-display-none');
-        printButton.removeAttribute('aria-hidden');
-        printButton.onclick = function () { print(); }
       }
+    }
+
+    const printAllButton = document.getElementById('print-all-button')
+    if (printAllButton) {
+      printAllButton.onclick = () => {
+        togglePrintBlock('csip-referral', true, false)
```
iterative-it, line 37: Perhaps to make it clearer what the result of each of these calls are doing, there could be a displayForPrint and hideFromPrint method that calls the underlying togglePrintBlock function, Also wonder if that could return a function pointer for displayForPrint('csip-investigation').andAddPageBreak() or just have displayForPrint and displayForPrintAndAddPageBreak separately. hideFromPrint could also take an array. Then you end up with just

displayForPrint('csip-referral');
hideFromPrint(['csip-investigation', 'csip-plan', 'csip-reviews'])

For the print referral onclick for example

iterative-it (approved):
LGTM

iterative-it (approved):
LGTM

PR #468, opened 2025-06-05
MIA-606: Use aria-live to callout conditional content

JackReeveMoJ:
Conditional content announces itself to screen readers when it appears using `aria-live`. No visible changes are made

Kapture 2025-06-05 at 16 06 32


iterative-it (approved):
LGTM

PR #486, opened 2025-06-27
MIA-523: Change content from "Triggers" to "Risks and triggers"

JackReeveMoJ:
image

image

image

image

image
(Note: AC states the details text should remain unchanged)

image


iterative-it (approved):
LGTM

PR #497, opened 2025-07-18
MIA-610: Prisoner CSIP page

JackReeveMoJ:
image


server/routes/prisoner-csips/view.njk
```diff
@@ -0,0 +1,71 @@
...
+    head: [
+      { text: 'Name and prison number', key: 'name' },
+      { text: "Location", key: 'location' },
+      { text: "Referral date", key: 'referralDate' },
+      { text: "CSIP log code", key: 'logCode' },
+      { text: "Case Manager", key: 'caseManager' },
+      { text: "Next review date", key: 'nextReviewDate' },
+      { text: 'CSIP status', key: 'status' },
+      { text: '' }
+    ] | select() | convertToSortableColumns(sort or 'nextReviewDate,asc', hrefTemplate),
```
iterative-it, line 67: I think this is a different default from the controller that uses `name,asc`. Ordering by name is a bit pointless since it only ever displays one person. Same with location. I'd drop sorting on both columns and change the default to `referralDate:desc`. We don't have clear requirements here but showing newest to oldest as a default seems reasonable

iterative-it (approved):
LGTM

PR #616, opened 2026-06-04
Run integration tests in parallel again

theDustRoom:
I had to remove the old parallel integration tests workflow as it used 3rd party code, this caused the tests to take approximately 30 minutes to run.
This reintroduces parallel integration tests with a simple script to split the tests instead of a 3rd party solution.

### DIFF
--- a/.github/workflows/parallel_integration_tests.yml
+++ b/.github/workflows/parallel_integration_tests.yml
@@ -43,7 +43,10 @@ jobs:
       - name: Get wiremock
         shell: bash
         run: |
-          curl -o wiremock.jar https://repo1.maven.org/maven2/org/wiremock/wiremock-standalone/3.9.1/wiremock-standalone-3.9.1.jar
+          WIREMOCK_VERSION="3.9.2"
+          WIREMOCK_SHA256="77b88dc41c3268e3ce391f3fef222b0cd0696f286d3fc35a84ba22ffcb877012"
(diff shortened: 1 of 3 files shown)


.github/workflows/parallel_integration_tests.yml
```diff
@@ -0,0 +1,85 @@
...
+          cache-name: node-modules
+        with:
+          path: |
+            ./node_modules
+            ~/.cache/Cypress
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+      - name: Get wiremock
+        shell: bash
+        run: |
+          curl -o wiremock.jar https://repo1.maven.org/maven2/org/wiremock/wiremock-standalone/3.9.1/wiremock-standalone-3.9.1.jar
```
xoen: maybe I'm a bit paranoid, but is it worth checking the checksum of this file before executing it? 🤔 

also, it looks like [3.9.2 is there as well](https://repo1.maven.org/maven2/org/wiremock/wiremock-standalone/3.9.2/).

xoen (dismissed):
FWIW this makes sense to me, just left a tiny comment ☺️

.github/workflows/parallel_integration_tests.yml
theDustRoom: No, that's a good catch.

xoen (approved):

